Issue #5, opened 2017-06-23
Mapping: Period<->Representative period


ckaldemeyer:
Hi again,

how can I obtain information about which periods have been selected e.g. weeks 2, 15, 35, and 46 are the chosen ones with the corresponding weighting factors (if any)?

Sorry for asking so many questions at once..

Best
Cord

ckaldemeyer:
And does the algorithm choose the same time slice for all the series e.g. a week with linked demand, wind and pv values or different weeks within each series e.g. week 1 and 43 for demand, 4 and 9 for wind and so on?

ckaldemeyer:
I think the first is the case, isn't it?

ckaldemeyer:
I found:

    # The order of the cluster Periods related to the original profile.
    tsam.TimeSeriesAggregation.clusterOrder

in the code.

I guess this should do it!

ckaldemeyer:
It works now. Thanks anyway!

Just an idea: would it make sense to store the chosen/representing cluster period in the predicted data?

Then, all information is combined and with a datetimeindex one could also group and analyse which clusters have been chosen for certain positions, etc.

You can close this issue anyway!

ckaldemeyer:
I have found the weights in the attribute `clusterPeriodNoOccur`so it works now. Maybe some general stats would be nice

l-kotzur:
Do you have any ideas how to improve the useability of the mapping information in the code?

ckaldemeyer:
> I have found the weights in the attribute clusterPeriodNoOccurso it works now. Maybe some general stats would be nice

Actually that only solved my problem partly and I still need to know which periods of the original time series have been chosen to represent the respective clusters.

My actual idea would be to go with pandas (like JOIN in SQL) and compare where the original an predicted data equals. The equal parts in the original data should be the selected periods, right?

> Do you have any ideas how to improve the useability of the mapping information in the code?

You mean generally like a reporting file?

The snipped I have is quite generic (I guess) and we might discuss along the code (pull request?) if parts could be integrated. But it's not something special so don't expect too much..

l-kotzur:
> Actually that only solved my problem partly and I still need to know which periods of the original time series have been chosen to represent the respective clusters.

This would be possible, but only for k-medoid and hierarchical clustering. Therefore an easy solution could be possible. Wherefore do you need the information?

> You mean generally like a reporting file?

[...]

PR #12, opened 2018-04-05
Get datetimeindex for hierarchical clustering

kimvk:
Referring to #5 it would be nice to receive information about the original time index (datetimeindex) of the clustered days, when using hierarchical clustering. 

We did creat a [fork of tsam](https://github.com/openego/tsam) and added a new functionality 'clusterCentersIndices', which gives us the indices of the cluster centers (which are used as medoids). 

The example  ['Add scrict example' ](https://github.com/openego/tsam/commit/c653c43bf1b877c674fe1e71b9c171f35017ecd5) shows how it can be applied to receive the corresponding datetimeindex of the original data. 

Could you check out this new functionality? Would it be possible to implement it in the original tsam repository?

### DIFF
--- a/example/script_example.py
+++ b/example/script_example.py
@@ -1,0 +1,25 @@
+import tsam.timeseriesaggregation as tsam
+import pandas as pd
+
+raw = pd.read_csv('testdata.csv', index_col=0)
+
+aggregation = tsam.TimeSeriesAggregation(raw, noTypicalPeriods = 8,
+                                         hoursPerPeriod = 24,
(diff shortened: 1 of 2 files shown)


ulfmueller:
@l-kotzur What do you think? Is this PR alright and sth. you would like to include?
Then we can get rid of our our fork and use directly the orginal tsam again...

l-kotzur:
I have integrated it into the master. Thanks for your contribution!

kimvk:
Thank you, that is great!

PR #15, opened 2018-05-30
Allow solver as optional argument

ckaldemeyer:
This implements the possibility to choose the solver for the k-medoids clustering algorithm from pyomo's `SolverFactory`.

Unfortunately, I could not test it because neither `pip install -e /path/to/local/tsam` nor `python setup.py install` is working on my system because it gets stuck (I have made sure that the pypi version had been deinstalled before). Might this be related to the setup.py? Or it is some weird behaviour of my system..

@l-kotzur: Could you check if it works and is sufficient like this?

### DIFF
--- a/tsam/timeseriesaggregation.py
+++ b/tsam/timeseriesaggregation.py
@@ -744,7 +744,7 @@ class TimeSeriesAggregation(object):
                 aggregatePeriods(
                     sortedClusterValues, n_clusters=self.noTypicalPeriods,
                     n_iter=30, solver=self.solver,
-                    clusterMethod=self.clusterMethod)
+                    clusterMethod=self.clusterMethod))
 
             clusterCenters_C = []
             distanceMedoid_C = []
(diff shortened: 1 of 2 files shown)


ckaldemeyer:
The API would be:

    aggregation = tsam.TimeSeriesAggregation(
      raw, noTypicalPeriods = 8, hoursPerPeriod = 24,
      clusterMethod = 'k_medoids', solver='gurobi')

l-kotzur:
Thanks already!

Can you add the solver argument also for the call of 'aggregatePeriods' in line 821? Otherwise it is only used for the sorted time series. I will further update the develop branch, and can you make the pull request to it instead of the master?

ckaldemeyer:
> Can you add the solver argument also for the call of 'aggregatePeriods' in line 821?

Done.

> I will further update the develop branch, and can you make the pull request to it instead of the master?

You can easily to it yourself by changing the base branch in the PR. But I can also do it if you have updated the dev. I guess the easiest thing would be to merge the master into the dev and then set the base. But you know better!

l-kotzur:
Thanks! I am also still learning collaborative development. 

I will test it, and pull it with a new version to the master.

ckaldemeyer:
> Thanks! I am also still learning collaborative development.
> 
> I will test it, and pull it with a new version to the master.

:+1:

PR #29, opened 2019-06-18
Predefining sequence update

maximilian-hoffmann:
@l-kotzur @l-welder possibility to determine new cluster medians with given cluster sequence only added for hierarchical clustering.


l-kotzur:
See #31

PR #30, opened 2019-06-18
Vectorization & MultiIndices

samuelduchesne:
Hi guys,
I love this package of yours and would like to contribute two little changes that improved my workflow while using TSAM.

First of all, in terms of speed, because of the general use in this package of loops to build DataFrames, it seems that some methods suffer with speed problems. Many [authors](https://engineering.upside.com/a-beginners-guide-to-optimizing-pandas-code-for-speed-c09ef2c6a4d6) suggest to use vectorization which can drastically improve some workflows. I added such an example in `_normalizeTimeSeries` by using the sklearn vectorized [min_max_scaler](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.MinMaxScaler.html) function. I suggest you try to gradually remove all the loops and replace them with vectorized version whenever possible :)

Second, I encountered an issue when there are column MultiIndexes in the original time series DataFrame. The error happens in `_preProcessTimeSeries` and is due to the `sorted()` method that breaks the column MultiIndex. The simplest fix is to use the [sort_index](https://pandas.pydata.org/pandas-docs/version/0.23.4/generated/pandas.DataFrame.sort_index.html#pandas.DataFrame.sort_index) method which can handle MultIndexes.

Best,

Sam

### DIFF
--- a/tsam/timeseriesaggregation.py
+++ b/tsam/timeseriesaggregation.py
@@ -414,19 +414,15 @@ class TimeSeriesAggregation(object):
         ---------
         normalized time series
         '''
-        normalizedTimeSeries = pd.DataFrame()
-        for column in self.timeSeries:
-            if not self.timeSeries[column].max() == self.timeSeries[column].min():  # ignore constant timeseries
-                normalizedTimeSeries[column] = (
-                                                   self.timeSeries[column] -
-                                                   self.timeSeries[column].min()) / \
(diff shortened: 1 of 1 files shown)


l-kotzur:
Hey Samuel,

thanks a lot for your contributions! I will create a feature branch and take a look.

Best,
Leander

PR #33, opened 2019-06-25
Vectorization

l-kotzur:
@samuelduchesne

Thanks again. I had to create a small work around to correct the normalization process for the sameMean case. Further, sklearn has some numerical tolerances in k-means which created negative values for zero filled time series in combination with the new normalization process. I also tried to correct this.

@maximilian-hoffmann 

Could you shortly review the new structure?


samuelduchesne:
Nice!
It could be interesting to measure the speed gain of using vectorized methods.

PR #39, opened 2019-09-12
Fixes an issue where the _addExtremePeriods method would fail due to MultiIndexes

samuelduchesne:
Hi @l-kotzur, when a MultiIndex is passed as columns, a tuple is passed as the column name for ExtremePeriods. This raises: "Can only concatenate tuple (not "str") to tuple error" since a string cannot be appended to a tuple. This fix fill append the extreme-period identifier to the last element of the tuple (lowest level).


samuelduchesne:
Also, this is completely unrelated, but may I suggest you guys run [black](https://pypi.org/project/black/) on the code base. I have been using black for a while and I absolutely recommend it. It basically formats the code without compromise. This helps a lot when creating pull requests since the formatting will not produce more diffs than it needs to.

maximilian-hoffmann:
Thank you very much for the change and the correction of "daily min" to "daily max" where needed.
Kind regards,
Max

samuelduchesne:
My pleasure!

PR #55, opened 2021-09-17
K-medoid contiguity

l-kotzur:
I have added a first version of the k-medoid contiguity algorithm.

@Shruti-Patil: Following code should work:

```python
from tsam.utils.k_medoids_contiguity import k_medoids_contiguity
n_clusters = 10
r_y, r_x, r_obj = k_medoids_contiguity(distance_matrix, n_clusters , adjacency_matrix, solver="gurobi")
labels_raw = r_x.argmax(axis=0)
```

There is some cleanup to do, since I would like to structure the k-medoids-contiguity class equivalent to the k-medoids class. I will do this, if I have time...

@OfficialCodexplosive: Can you make a new tsam-Version 1.1.2 in pip, such we can directly use it?

@maximilian-hoffmann: Sorry for abusing tsam for this, but the existing exact k-medoid algorithm just gave a good starting point. If it is not wanted here in future, we will move it. Could you approve a merge and then forward to Julian?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # directories
 .idea/
+.vscode/
 __pycache__/
 trash/
 *.pytest_cache/
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,3 +2,4 @@ sklearn>=0.0
 pandas>=0.18.1
 numpy>=1.11.0
 pyomo>=5.3
+networkx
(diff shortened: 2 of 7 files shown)


maximilian-hoffmann (approved):

maximilian-hoffmann:
We'll probably get a merge conflict once I will merge the duration curve representation, but @OfficialCodexplosive you can proceed with setting up version 1.1.2 as @l-kotzur said.

PR #63, opened 2022-05-01
Segment period combo

l-kotzur:


maximilian-hoffmann (changes requested):
hyperparametertuning.py:

- for identifyOptimalSegmentPeriodCombination(), also implement it the other way round: Starting from minSegment-Number, increase Segments while decreasing number of periods. We'll end up at a second aggregation-proposal. Take the better out of the two. Alternatively: Check all combinations until maxPeriods is reached? Are you sure, that RMSE as function of the period and segment numbers you are checking has only one minimum? Im not perfectly sure here for hierarchical clustering. For other aggregations and in order to improve the robustness of the algorithm, I wouldn't use this convergence criterion and simply check all combinations.

- I don't get the purpose of the second and third while loop of the pareto optimal aggregation in line 215 ff. Please explain.

Let's make an appointment during working times when you would like to give an in-person explanation.
Kind regards,
Max

l-kotzur:
Hey Max, thanks for the hints.

Could you correct the first point or add your code?

Regarding the second point: Second and third while loop are required since the first loop has an "and" condition and stops in case a maximum of periods or segments is reached. E.g., the while loop would stop in case the maximum segment number is reached but the periods number is still below the the maximum. Therefore, the other loops are required. 

Basically, the first while loop goes through the open solution space, and the second and third along the borders. I am pretty sure there are better implementations... Please feel free to improve in future. But I guess for a first release, this is fine.

Best
Leander

l-kotzur:
@maximilian-hoffmann Could you approve?

maximilian-hoffmann (approved):

PR #65, opened 2022-05-29
Include distribution representation with min max and a small bug fix in the later

l-kotzur:
Release 2.1.0


l-kotzur:
@maximilian-hoffmann tbn

PR #67, opened 2022-11-28
Update README.md according to institute standard

OfficialCodexplosive:


l-kotzur:
The README has updated to many times and the majority of changes are already included.

PR #72, opened 2023-06-28
Fix pandas issue, fix sum for distribution representation inlc. min max vals, add possibility to define segment representation and update the default example

l-kotzur:


l-kotzur:
@rschwarz Somehow one of the tests fails with the new pandas version: https://github.com/FZJ-IEK3-VSA/tsam/actions/runs/5401641896/jobs/9811733821

We will fix it and then release it. 

@maximilian-hoffmann Can you take a look at this? The branch with the isolated changes due to the new pandas version > 2.0.0 is here: https://github.com/FZJ-IEK3-VSA/tsam/tree/pandas_fix 

I cannot figure out why the results change for the segmentation... although it are only small changes. If you feel comfortable with it, we can adapt the results.

rschwarz:
Possible some of the reformulated pandas expressions are not actually equivalent to what was there before. I followed the general pattern of suggested changes in other code bases and the hints given by the deprecation warning. But, lacking an understanding of the code in context, might have introduced a different behavior.

I _believe_ that I did run the tests successfully locally, but it's been a while...

l-kotzur:
I think it is a problem of the test itself.

l-kotzur:
New release 2.3.0

@rschwarz Thanks again for finding and fixing the bug.

PR #78, opened 2023-08-21
setup.py: correct current version

fneum:


l-kotzur:
@fneum Thanks, but I saw this already and solved it with the latest documentation push.

PR #79, opened 2023-08-21
Performance improvement rescaleClusterPeriods

l-kotzur:
@fneum tbn


l-kotzur:
@fneum One thing to point out: If you use the 'distributionRepresentation', you can turn off rescaleClusterPeriods since the total sum (or mean) of the time series values are kept intrinsically.

PR #82, opened 2024-01-25
Small improvements to timeseriesaggregation

ddceruti:
Fixed a Pandas FutureWarning, improved slightly two exceptions raised so that the complete traceback is preserved and deleted a unused import.

```Python
tsam\timeseriesaggregation.py:875: FutureWarning: A value is trying to be set on a copy of a DataFrame or Series through chained assignment using an 
inplace method.
The behavior will change in pandas 3.0. This inplace method will never work because the intermediate object on which we are setting values always behaves as a copy.

For example, when doing 'df[col].method(value, inplace=True)', try using 'df.method({col: value}, inplace=True)' or df[col] = df[col].method(value) instead, to perform the operation inplace on the original object.
```

### DIFF
--- a/tsam/timeseriesaggregation.py
+++ b/tsam/timeseriesaggregation.py
@@ -870,9 +870,9 @@ class TimeSeriesAggregation(object):
                 )
 
                 # reset values higher than the upper sacle or less than zero
-                typicalPeriods[column].clip(lower=0, upper=scale_ub, inplace=True)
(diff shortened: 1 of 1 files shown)


l-kotzur:
@ddceruti Thanks a lot for the update and the merge request.

@adbuerger mentioned further depreciation warnings with pandas 3.0. I will fix these together with this one.

Issue #81, opened 2024-01-25
Expected incompatibility with pandas 3.0 due to changes in stack function implemetation


adbuerger:
Dear tsam developers,

when using a recent version of pandas, I receive the following deprecation warning when using tsam:

`
The previous implementation of stack is deprecated and will be removed in a future version of pandas. See the What's New notes for pandas 2.1.0 for details. Specify future_stack=True to adopt the new implementation and silence this warning.
`

The notes referenced in the deprecation warning state that from pandas 3.0 on, the previous implementation of the stack-function will not be available anymore, cf. https://pandas.pydata.org/docs/whatsnew/v2.1.0.html#new-implementation-of-dataframe-stack. According to their milestones, release of pandas 3.0 could be expected already around April 2024, cf. https://github.com/pandas-dev/pandas/milestone/102.

In tsam, the mentioned `stack` function is used in multiple positions in `timeseriesaggregation.py` and `durationRepresentation.py`. Since, e.g., the default sorting behavior is expected to change in the new implementation, I am unsure whether these changes in behavior could impact the outputs of tsam. Therefore, I wanted to ask whether you are already aware of this change and if it is already planned to adjust the use of pandas at these points in the code, accordingly.

Thank you for providing this very useful software!

julian-belina:
@adbuerger Thank you for pointing out the compatability issue and sorry for the long wait. Version 2.3.3 is now compatabile with the newer pandas and numpy versions. V2.3.2 is compatible with older dependencies.

PR #83, opened 2024-08-08
Limit pandas

jo-omoyele:
Limit pandas and Numpy. File deprecation warnings

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -12,14 +12,20 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
-
+    runs-on: ${{matrix.os}}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ["ubuntu-latest","ubuntu-20.04", "macos-latest","macos-13","macos-12", "windows-latest","windows-2019"]
+        # os: ["ubuntu-latest"]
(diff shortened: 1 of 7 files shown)


julian-belina (approved):
Looks good to me

l-kotzur:
Heyho @jo-omoyele @julian-belina 

Thanks for contributing!

Could you shortly explain why you did this? This limits the usage for. Is it just about the depreciation warning or is it already failing? If the first, it would be better to fix the code instead of limiting the version. If you do the latter, you will destroy the integration of the whole user base using newer versions of pandas or Python. 

Please give a short hint before releasing on PyPI. I can also take a look.

julian-belina:
Hey @l-kotzur, we did some testing and the tests are already failing for the exlcuded versions. We are planning to update the code so its compatible with the new versions of its dependencies. Unfortunately I cant give a timeline yet.

l-kotzur:
Ok. Perhaps I will find some time later this day and can take a look. There is an issue for this #81 and a pull request #82 doing parts of this fix.

jo-omoyele:
Hello @l-kotzur, we are aware of these issues. It might take a couple of weeks to fix properly based on our schedules. However, we needed a quick approach to filter the warnings while we take our time fixing the issues.

julian-belina:
@l-kotzur, thank you any help is appreciated.

Issue #90, opened 2024-09-18
invalid escape sequence '\('


julian-belina:
When running the test of tsam I get the following warnings associated with escape sequences:

> test\test_assert_raises.py:21: 10 warnings
>   C:\Programming\tsam\test\test_assert_raises.py:21: SyntaxWarning: invalid escape sequence '\('
>     "timeSeries has to be of type pandas.DataFrame\(\) or of type np.array\(\) in "
> 
> test\test_assert_raises.py:129: 10 warnings
>   C:\Programming\tsam\test\test_assert_raises.py:129: SyntaxWarning: invalid escape sequence '\['
>     "clusterMethod needs to be one of the following: \['averaging', 'k_means', "
> 
> test\test_assert_raises.py:130: 10 warnings
>   C:\Programming\tsam\test\test_assert_raises.py:130: SyntaxWarning: invalid escape sequence '\]'
>     "'k_medoids', 'k_maxoids', 'hierarchical', 'adjacent_periods'\]",
> 
> test\test_assert_raises.py:147: 10 warnings
>   C:\Programming\tsam\test\test_assert_raises.py:147: SyntaxWarning: invalid escape sequence '\['
>     "extremePeriodMethod needs to be one of the following: \['None', 'append', "
> 
> test\test_assert_raises.py:148: 10 warnings
>   C:\Programming\tsam\test\test_assert_raises.py:148: SyntaxWarning: invalid escape sequence '\]'
>     "'new_cluster_center', 'replace_cluster_center'\]",

It occurs on python 3.9 with numpy 2.0.2 but not on python 3.12

arnavsharma990:
The issue is caused by invalid escape sequences in string literals

jo-omoyele:
Issue resolved

PR #97, opened 2025-05-28
Python update

jo-omoyele:
os update:

- Added Python 3.13 to the action matrix
- Updated the Python requirement in the setup.py to incorporate 3.13
- Updated ubuntu-20.4 to ubuntu-22.04 in workflow due to deprecation
- Removed macos-12 due to slow runs
- Solved issue #90


julian-belina (changes requested):
The version number in the pyproject.toml needs to be incremented

julian-belina (approved):
looks good to me

Issue #96, opened 2025-05-06
time series aggregation with segmentation is considerably slower than without


phil-fzj:
If one is using segmentation when doing time-series aggregation the processing takes significantly longer compared to doing it without segementation.
I dug a little deeper and benchmarked the steps of the timeseries aggregation.
I noticed that there is a large loop where pandas dataframes are stacked:
https://github.com/FZJ-IEK3-VSA/tsam/blob/c46fbd3e577dc823d79cc688cfef671603c8576d/src/tsam/utils/durationRepresentation.py#L60

This is responsible for a large chunk of the time that tsam takes when using segmentation.
I think an implementation that does not use the .stack() would greatly improve the processing time of tsam.

PR #98, opened 2025-06-10
Fix long calculation times when using duration representation.

phil-fzj:
Adresses #96 
 Now using numpy operations where possible, minimizing DataFrame manipulations when using duration representation.


julian-belina (approved):
Looks good to me

Issue #99, opened 2025-06-13
Aggregation results differ between 2.3.7 and 2.3.8


adbuerger:
I just upgraded tsam from 2.3.7 to 2.3.8 and I see that the aggregation results for the same input data have changed though I did not change any settings on my side. Since my application uses duration representation I assume this is related to the changes #98

Here are some sample outputs:

Before (2.3.7):

![Image](https://github.com/user-attachments/assets/14a05ea1-e9a3-4722-b071-49fcedeba50b)

After (2.3.8):

![Image](https://github.com/user-attachments/assets/fb2d5579-bfdf-4b77-9cc3-e9e2a5ef1229)

julian-belina:
@adbuerger The issue should be resolved in version 2.3.9. Could you please confirm this? The PyPI package has been uploaded and the Conda Forge package should be ready to download in approximately two hours.

julian-belina:
You're welcome! The credit for the speed improvement goes to @phil-fzj. Thank you also for reporting the issue.

PR #100, opened 2025-06-13
revert and fix previous changes to durationRepresentation 

phil-fzj:
revert previous changes to durationRepresentation as they were producing wrong results as indicated by #99. 
reimplemented the original fix. It should also be much cleaner now.
Only really necessary code parts are changed compared to working version in v.2.3.7
(compare here: https://github.com/FZJ-IEK3-VSA/tsam/compare/v.2.3.7...phil-fzj:tsam:96_fix_slow_duration_representation_99_fix)
added test that compares clustering against clustered values based on v2.3.7.


phil-fzj:
I had to remove the newly crated test due to the issue specified in #101
However, on the same machine, os and python version the changed function now produces the same outupt as v2.3.7

PR #117, opened 2025-10-30
108 update read the doc

jo-omoyele:
The read the doc is built. Sphinx and conda forge are added for the build up


julian-belina (approved):
looks good to me

PR #146, opened 2026-02-06
feat: missing representation objects, bugfixes and new tests

FBumann:
## Summary

Adds a comprehensive regression test suite and benchmark infrastructure for tsam v3, ensuring correctness against v2.3.9 baselines and providing tooling for performance tracking across versions, along with some bugfixes in v3.0.0 and missing features in 3.0.0

### New Features

- **`Distribution` and `MinMaxMean` representation objects** for `ClusterConfig` and `SegmentConfig`, providing a structured alternative to plain string representation names
- **Migration guide** (`docs/source/migrationGuideDoc.rst`) documenting the transition from the v2 class-based API to the v3 functional API

### Bug Fixes

- **Segment label ordering bug**: `AgglomerativeClustering` produces arbitrary cluster labels, which caused `durationRepresentation()` with `distributionPeriodWise=False` to allocate the global distribution differently when transferring a clustering. Segment clusters are now relabelled to temporal order after `fit_predict()`.
- **Fixed non-deterministic sorting in `durationRepresentation()`** across both code paths. The `distributionPeriodWise=True` path now uses vectorized numpy 3D operations with `kind="stable"` and `np.round(mean, 10)` before `argsort` to ensure identical tie-breaking across platforms. The `distributionPeriodWise=False` path applies the same stable sort and rounding fixes.
- **Regenerated golden files and `test_clustering_e2e` expected files** to match the numpy+stable implementation.

### Testing & Benchmarks

- **Shared test configs** (`test/_configs.py`) — single source of truth for all 27 algorithm configurations, 4 datasets, and cross-product case building. No new-API imports, so it works under any tsam version.
- **Old/new API equivalence tests** (`test_old_new_equivalence.py`) — verifies `TimeSeriesAggregation` and `tsam.aggregate()` produce identical results: 27 configs × 4 datasets = 74 cases × 4 comparisons = **296 tests**
- **Golden-file regression tests** (`test_golden_regression.py`) — compares both APIs against stored CSV baselines generated with tsam v2.3.9: **148 tests**
- **Benchmark suite** (`benchmarks/bench.py`) — for comparing performance across versions using pytest-benchmark
- Replaced OPSD dataset with derived `wide.csv` (12 columns, 3 countries) to avoid third-party data licensing concerns

### Architecture

All test configurations are defined once in `test/_configs.py`:

```
test/_configs.py                    # 27 BaseConfigs (old_kwargs only), datasets, helpers
├── test/test_old_new_equivalence.py  # Adds new_kwargs, runs old == new comparison
│   └── test/test_golden_regression.py  # Imports cases, compares against stored CSVs
├── test/generate_golden.py           # Generates golden CSVs (runs under v2.3.9)
└── benchmarks/bench.py               # Performance benchmarks (runs under any version)
```

`_configs.py` deliberately avoids importing any v3 API types (`ClusterConfig`, etc.), so it can be imported even when tsam v2.3.9 is installed. This is what enables `generate_golden.py` and `benchmarks/bench.py` to work against old versions.

### Datasets

| Dataset | Description | Source |
|---|---|---|
| `testdata` | 4 columns (GHI, T, Wind, Load), 8760 hours | Existing example notebook data (MIT) |
| `wide` | 12 columns (DE/FR/NL × GHI/T/Wind/Load), 8760 hours | Derived from testdata with scaling, time shifts, seeded noise |
| `constant` | 2 constant columns, 10 days | Synthetic — tests edge case where all periods are identical |
| `with_zero_column` | testdata + all-zero column | Tests zero-variance edge case |

### Coverage

The 27 configurations cover:

- **6 cluster methods**: hierarchical, k_means, k_medoids, k_maxoids, contiguous (adjacent_periods), averaging
- **7 representation types**: medoid (default), mean, maxoid, distribution, distribution+minmax, distribution (global scope), minmax_mean
- **3 extreme methods**: append, replace_cluster_center, new_cluster_center
- **4 extreme types**: max_value (addPeakMax), min_value (addPeakMin), max_period (addMeanMax), min_period (addMeanMin)
- **Options**: segmentation, no rescaling, weighted columns, combined extremes+segmentation

### Test layers

1. **Equivalence tests** (`test_old_new_equivalence.py`): For each of 74 config×dataset cases, verifies:
   - `cluster_representatives` DataFrames are identical
   - `cluster_assignments` arrays are identical
   - `accuracy` metrics (RMSE, MAE) match within tolerance
   - `reconstructed` time series are identical

2. **Golden regression tests** (`test_golden_regression.py`): For each of 74 cases, verifies:
   - New API `reconstructed` matches stored v2.3.9 golden CSV
   - Old API `predictOriginalData()` matches stored v2.3.9 golden CSV

   This catches regressions where both APIs change together (which equivalence tests alone cannot detect).

3. **Golden file generation** (`generate_golden.py`): Run against a specific tsam version to produce baselines:
   ```bash
   uv run --with tsam==2.3.9 --no-project test/generate_golden.py
   ```

## Benchmarking workflow

The benchmark suite lives in `benchmarks/bench.py` (outside `test/`, not collected by CI). It uses the same 74 config×dataset cases from `_configs.py` and only calls the old `TimeSeriesAggregation` API, so it runs against any tsam version.

### Comparing versions

```bash
# 1. Benchmark old version
uv pip install tsam==2.3.9
pytest benchmarks/bench.py --benchmark-save=v2.3.9

# 2. Benchmark current version
uv pip install -e .
pytest benchmarks/bench.py --benchmark-save=v3.1.1

# 3. Compare
pytest benchmarks/bench.py --benchmark-compare='*v2.3.9'
# or
pytest-benchmark compare '*v2.3.9' '*v3.1.1' --group-by=name
```

Results are saved to `.benchmarks/` (gitignored, machine-specific).

### Performance results (current vs v2.3.9)

| Config | constant | testdata | wide | with_zero_column |
|---|---|---|---|---|
| **hierarchical_default** | 2.3x | 44.4x | 25.3x | 42.0x |
| **hierarchical_mean** | 4.4x | 50.4x | 37.7x | 49.0x |
| **hierarchical_maxoid** | 4.8x | 56.7x | 40.3x | 54.0x |
| **hierarchical_distribution** | 5.0x | **54.5x** | **35.1x** | **50.8x** |
| **hierarchical_dist_minmax** | 5.1x | **55.1x** | **37.5x** | **52.4x** |
| **hierarchical_no_rescale** | 4.8x | 59.0x | 45.9x | 55.4x |
| **hierarchical_segmentation** | 3.4x | 34.6x | 28.6x | 32.7x |
| **hierarchical_weighted** | — | 55.5x | — | — |
| **contiguous** | 4.7x | 54.3x | 49.9x | 53.2x |
| **averaging** | 5.0x | **76.9x** | **66.2x** | **73.9x** |
| **kmeans** | 1.4x | 4.3x | 5.5x | 6.3x |
| **kmaxoids** | 1.3x | 1.4x | 1.4x | 1.4x |
| **kmedoids** | — | 0.9x | — | — |
| **distribution_global** | 2.0x | 15.5x | 6.7x | 13.3x |
| **distribution_minmax_global** | 1.9x | 14.8x | 6.6x | 13.2x |
| **minmaxmean** | — | 63.2x | — | — |
| **extremes_*** | 4.7x | 32–58x | 41.2x | 31–54x |

- Hierarchical methods on real-world data: **35–60x faster**
- **Averaging** biggest winner: 77x (590ms → 8ms)
- **Distribution configs** (cluster-wise): **35–55x faster** (numpy vectorization of `durationRepresentation`)
- Distribution configs (global scope): **7–16x faster**
- Iterative methods (kmeans, kmedoids, kmaxoids): 1–6x (core solver dominates)

## Test plan

- [x] `pytest test/test_old_new_equivalence.py` — 296 tests pass
- [x] `pytest test/test_golden_regression.py` — 148 tests pass (74 new API + 74 old API match v2.3.9 golden baselines)
- [x] `pytest test/test_clustering_e2e.py` — all tests pass (expected files regenerated for pandas implementation)
- [x] Benchmarked all 74 cases against v2.3.9


FBumann:
@julian-belina As we see from the failing CI: **The pandas revert didn't help** — even v2.3.9 uses np.sort() without kind="stable", so it was never cross-platform deterministic.
Going with PR #147 (numpy vectorized + kind="stable") — it's the only option that's deterministic (and fast).

FBumann:
@julian-belina This PR is final.

benchmarks/bench.py
```diff
@@ -0,0 +1,63 @@
+"""Benchmarks for tsam using pytest-benchmark.
+
+Self-contained: only imports ``tsam.timeseriesaggregation`` so it works with
+both the current dev version and older releases (e.g. tsam==2.3.9).
+
+Usage::
+
+    # Benchmark an old version
+    uv pip install tsam==2.3.9
+    pytest benchmarks/bench.py --benchmark-save=v2.3.9
```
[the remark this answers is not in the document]
julian-belina: Instead of following the description here, I created a new environment with TSM 2.3.9 using Conda. I then received the following error:

```
________________________________________________ ERROR collecting benchmarks/benchmark_tuning.py _________________________________________________
ImportError while importing test module '/fast/home/j-belina/tsam_bumann/tsam/benchmarks/benchmark_tuning.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
../../miniforge3/envs/tsam_env_2_3_9/lib/python3.13/importlib/__init__.py:88: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
benchmarks/benchmark_tuning.py:24: in <module>
    from tsam.tuning import find_optimal_combination
E   ModuleNotFoundError: No module named 'tsam.tuning
```

The problem is that tsam 2.3.9 does not include the tuning module in the 'benchmark_tuning.py' file. 

`from tsam.tuning import find_optimal_combination`

Perhaps the best approach is to use the subprocess module, meaning that no Python modules need to be imported in the first module. This is just an idea, though.

src/tsam/utils/durationRepresentation.py
```diff
@@ -102,7 +90,7 @@ def durationRepresentation(
                 # get all the values of a certain attribute and cluster
                 candidateValues = candidates_df.loc[indice[0], a]
                 # calculate centroid of each cluster and append to list
-                meanVals.append(candidateValues.mean())
+                meanVals.append(np.round(candidateValues.mean(), 10))
```
julian-belina: Why do we need this here?

src/tsam/utils/durationRepresentation.py
```diff
@@ -42,54 +42,42 @@ def durationRepresentation(
...
-            )  # (n_attrs, timesteps)
-
-            # Reorder repr_values according to orders
+                repr_values[:, 0] = flat[:, 0]
+                repr_values[:, -1] = flat[:, -1]
+
+            # Reorder each attribute's repr_values by its mean profile.
+            # Round means before argsort to ensure identical tie-breaking
+            # across platforms and numpy versions.
+            means = np.round(cluster_data.mean(axis=1), 10)
```
julian-belina: Same here — why are we rounding? Is this to ensure consistent test results? If possible, I think it would be better to increase the test tolerance than start rounding during aggregation.

test/test_api_equivalence.py
```diff
@@ -327,6 +335,200 @@ def test_distribution_minmax_representation(self, sample_data):
...
+            n_clusters=8,
+            preserve_column_means=False,
+            cluster=ClusterConfig(
+                method="hierarchical",
+                representation=Distribution(scope="global"),
+            ),
+        )
+
+        pd.testing.assert_frame_equal(
+            old_result, new_result.cluster_representatives, check_names=False
```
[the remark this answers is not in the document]
julian-belina: If there are very small deviations between versions, you can use check_exact=False and adjust the tolerance accordingly. I think this is better than rounding during aggregation.

src/tsam/utils/durationRepresentation.py
FBumann: If we don't round, mean creates different results on different system architectures (sub 10e-15)
Without, sorting on means is not deterministic

julian-belina:
@FBumann This looks very promising, but I need to be able to run the tests/benchmarks on my machine. Then we can merge.

julian-belina:
Also I get the following warnings when running.

`pytest test/test_golden_regression.py -v`

The legacy warnings should be silenced when running Pytest. I am not sure if the ConvergenceWarning can be ignored, but it should be either silenced or addressed.
```
test/test_golden_regression.py::TestGoldenRegression::test_new_api_matches_golden[kmeans/constant]
test/test_golden_regression.py::TestGoldenRegression::test_old_api_matches_golden[kmeans/constant]
  /fast/home/j-belina/miniforge3/envs/tsam_3_1_1/lib/python3.14/site-packages/sklearn/base.py:1336: ConvergenceWarning: Number of distinct clusters (1) found smaller than n_clusters (3). Possibly due to duplicate points in X.
    return fit_method(estimator, *args, **kwargs)

test/test_golden_regression.py: 77 warnings
  /fast/home/j-belina/tsam_bumann/tsam/test/test_old_new_equivalence.py:301: LegacyAPIWarning: TimeSeriesAggregation is deprecated and will be removed in a future version. Use tsam.aggregate() instead. See the migration guide in the documentation.
    agg = old_tsam.TimeSeriesAggregation(timeSeries=data, **case.old_kwargs)
```

FBumann:
> @FBumann This looks very promising, but I need to be able to run the tests/benchmarks on my machine. Then we can merge. 

If you follow the instructions, does it still not work?
I'm not really firm in conda

benchmarks/bench.py
FBumann: Its not to be run with pytest!
Use `python benchmarks/benchmark_tuning.py`

Issue #150, opened 2026-02-07
Cross-platform determinism in durationRepresentation: comparison of approaches


FBumann:
## Background

The `durationRepresentation()` function in `src/tsam/utils/durationRepresentation.py` produces different results across platforms (macOS vs Ubuntu vs Windows). This was discovered via golden regression tests failing in CI.

**Root cause**: two operations are sensitive to platform/version differences:

1. **`np.sort()` with default `kind="quicksort"`** — numpy's introsort is *not* a stable sort. When values are equal (ties), the output order is platform-dependent. Both the v2.3.9 pandas implementation and the numpy rewrite use `np.sort()` without `kind="stable"`, so **both are affected**.
2. **`np.argsort()` / `.sort_values()` on `np.mean()` results** — floating-point accumulation order in `np.mean()` can differ by ~1e-16 depending on memory layout, SIMD optimizations, and numpy version. When means are nearly equal, this flips sort ordering.

**Both code paths are affected**: `distributionPeriodWise=True` and `distributionPeriodWise=False`.

**This is not a regression introduced by v3** — v2.3.9 was never cross-platform deterministic. It just wasn't tested across platforms.

### Core fixes

**1. Use stable sorting algorithms** (`kind="stable"` on all sort/argsort calls)
**2. Round mean values to eliminate 1e-16 machine level differences in sorting** (`np.round(mean, 10)`)

### Codebase audit

All other sort operations in the codebase were audited and are safe:
- `segmentation.py` — sorts on unique integer indices (no ties possible)
- `timeseriesaggregation.py:1043` — already uses `kind="stable"`
- `timeseriesaggregation.py:1332` — `np.sort(np.unique(...))` on unique integers
- `timeseriesaggregation.py:1511-1512` — sort for RMSE_duration metric (tie order doesn't affect result)
- `tuning.py` — Python `sort()` (already stable) and `sorted()` on unique integers
- `plot.py` / `result.py` — display/output only

## Base: PR #146 (includes #148)

PR #146 is the base branch. It includes:
- Extensive testing infrastructure (golden regression tests, e2e clustering tests, old↔new equivalence tests)
- Golden data produced against **v2.3.9**
- The v2.3.9 pandas implementation from PR #148 (already merged into #146)

This means #146 currently uses the original v2.3.9 pandas code and still **fails CI on Ubuntu/Windows** — because v2.3.9 was never cross-platform deterministic either.

## Approaches evaluated

[...]

PR #151, opened 2026-02-08
Fix/determinism pandas

FBumann:
## Summary

- Adds `kind="stable"` to all `np.sort()` / `.sort_values()` calls in `durationRepresentation.py` (both `distributionPeriodWise=True` and `=False` paths)
- Adds `np.round(mean, 10)` before sorting on mean values to absorb ~1e-16 floating-point accumulation differences
- Keeps the original v2.3.9 pandas-based code structure
- Regenerates golden and e2e fixture files

This is **Option A** from #150. See that issue for full context, benchmarks, and comparison with Option B (PR #152).

## Test plan

- [x] All 635 tests pass locally
- [ ] CI passes on all platforms (macOS, Ubuntu, Windows)


FBumann:
Closed in favor of #152

PR #175, opened 2026-03-14
Move warning suppression to test-specific markers

FBumann:
This PR removes all filterwarnings from pyproject from toml, as having them there filters them out for all users, effectively removing them from the library. I assume this change was made to not have warnings in tests. I added filterwarnings to the tests that raise them.
I would much rather keep them for users. Its somewhat a design decision, so if you really want to filter them out completely thats up to you

## Summary

- Remove blanket `filterwarnings` from `pyproject.toml` that were hiding tsam's own warnings (LegacyAPIWarning, ConvergenceWarning, edge-case UserWarnings) globally during test runs
- Add targeted suppression to each test file using `pytestmark`, `@pytest.mark.filterwarnings`, and `warnings.catch_warnings()` context managers
- Keep only third-party warnings (threadpoolctl RuntimeWarning, sklearn memory leak) suppressed globally since those are outside tsam's control

## Details

| Warning | Suppression location |
|---------|---------------------|
| `LegacyAPIWarning` | `pytestmark` on old-API test files + `catch_warnings` in `_run_old()` |
| `ConvergenceWarning` | Class marker on `TestOldNewEquivalence` + context manager in golden tests |
| `"At least one maximal value"` | Same as above (kmaxoids/wide, hierarchical_distribution_minmax/wide) |
| `"Segmentation is turned off"` | `pytestmark` on `test_properties` |
| `"The cluster is too small"` | `pytestmark` on `test_durationRepresentation` |
| `"Max iteration number reached"` | `pytestmark` on `test_reconstruct_samemean_segmentation` |

## Test plan

- [ ] `pytest test/ -q` passes with no unexpected warnings
- [ ] `pytest test/ -W error` passes (all expected warnings are suppressed at the right level)

### DIFF
--- a/src/tsam/plot.py
+++ b/src/tsam/plot.py
@@ -74,7 +74,7 @@ def _validate_columns(
     valid = [c for c in requested if c in available]
     invalid = [c for c in requested if c not in available]
 
-    if invalid:
+    if invalid and valid:
         warnings.warn(
             f"Columns not found in {context} and will be ignored: {invalid}. "
             f"Available columns: {available}",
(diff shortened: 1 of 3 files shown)


FBumann:
Good work!

PR #208, opened 2026-03-24
ci: fix docs build and add upper bounds to develop dependencies

### LINKED ISSUE #204: Renovate uses PyPI datasource instead of conda-forge for conda environment.yml packages
## Description

Renovate bot (PR #197) proposes updating `numpy` in `environment.yml` to `>=1.22.4,<=2.4.3`, which is correct. However, the confidence/age badges in that PR reveal that Renovate is using the **PyPI datasource** (`pypi/numpy/2.4.3`) to discover and validate new versions — not the conda-forge datasource — even though `environment.yml` explicitly lists `conda-forge` as the only channel.

## Root Cause

Renovate's `conda` manager does not automatically read the `channels:` list from `environment.yml` to determine which datasource to use for version lookups. Instead, it falls back to **PyPI** for packages that don't have an explicit datasource override configured.

This can be seen in the PR #197 badge URL:
```
https://developer.mend.io/api/mc/badges/age/pypi/numpy/2.4.3
```

The current `renovate.json` uses `"rangeStrategy": "widen"` but does not configure a `registryUrls` override for the `conda` manager to point at conda-forge.

## Why This Is a Problem

- Version availability on **PyPI** and **conda-forge** is not always in sync. A version could be on PyPI but not yet (or never) packaged for conda-forge, leading Renovate to propose an update that cannot actually be installed in the conda environment.
- Confidence and age metrics shown in PRs are based on PyPI adoption, not conda-forge adoption, giving a misleading signal for conda environments.

## Suggested Fix

Add a `registryUrls` override for the `conda` manager in `.github/renovate.json` to use the conda-forge datasource:

```json
{
  "packageRules": [
    {
      "matchManagers": ["conda"],
      "registryUrls": ["https://conda.anaconda.org/conda-forge"]
    }
  ]
}
```

This ensures Renovate queries conda-forge directly for version discovery, aligning with the channel declared in `environment.yml`.

## References

- PR demonstrating the issue: #197
- Renovate conda datasource docs: https://docs.renovatebot.com/modules/datasource/conda/

julian-belina:
## Summary

- Excludes `gen_ref_pages.py` from mkdocs-jupyter processing — it lives in `docs/` because mkdocs-gen-files requires scripts relative to the project root, but mkdocs-jupyter would otherwise try to execute it as a notebook, crashing the build
- Pins `mkdocs-jupyter<0.26` to avoid a regression in 0.26.0 where the notebook template fails to resolve `mkdocs_html/assets/index.css` via nbconvert
- Adds upper version bounds to all develop dependencies in `pyproject.toml` and `environment.yml` (set to current latest versions at time of commit) so future regressions like this are caught before they silently break CI — Renovate will keep them up-to-date via PRs

Closes #204

## Test plan

- [x] Verify docs build passes in CI


FBumann:
@julian-belina But thats usually not an issue with dev/docs dependencies. And no concern for **users** of tsam.
Im fine with upper bounds if you prefer.

julian-belina:
@FBumann, that would be much appreciated. Could you do the GitHub review? Otherwise, I cannot squash it.

PR #211, opened 2026-03-24
ci: create one Renovate PR per dependency instead of grouping all together

julian-belina:
## Summary

- Removes the `python-dependencies` group rule that bundled all package updates into a single PR (see #209 as an example of the problem)
- Adds a `groupName: "{{depName}}"` rule scoped to `pep621` and `custom.regex`, so each package (e.g. `numpy`) gets **one PR** that updates both `pyproject.toml` and `environment.yml` together
- Raises `prConcurrentLimit` to 20 and `prHourlyLimit` to 10 to accommodate the higher PR volume
- Follows the same approach used in [FINE](https://github.com/FZJ-IEK3-VSA/FINE/blob/develop/.github/renovate.json)

## Test plan

- [ ] Merge and wait for next Renovate run — verify individual per-package PRs are opened instead of one bulk PR

### DIFF
--- a/.github/workflows/push_to_jugit.yml
+++ b/.github/workflows/push_to_jugit.yml
@@ -1,0 +1,28 @@
+name: Mirror to JuGit
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - develop
+      - master
+
+jobs:
+  mirror:
+    name: Push ${{ github.ref_name }} to JuGit
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


julian-belina:
@FBumann This is almost my last pull request for today. The upper dependency limit check should work better now, enabling you to upgrade dependencies individually.

The next step is to update the dependency limits (except mkdocs-jupyter). Then we're ready for release from my side.

PR #234, opened 2026-03-26
feat!: new pipeline architecture

FBumann:
## Motivation

The v3 functional API (`tsam.aggregate()`) was a thin wrapper around the legacy `TimeSeriesAggregation` class — all the actual logic still lived in a 1,200-line monolith with deeply entangled state. This made bugs hard to trace, behavior hard to test in isolation, and the codebase hard to contribute to.

## What this PR does

**Architecture: pipeline decomposition**

`tsam.aggregate()` is now the single entry point. It builds a configuration and hands off to `run_pipeline()`, a stateless sequence of pure functions split across two layers:

| `src/tsam/pipeline/` | Responsibility |
|---|---|
| `orchestrator.py` | The four phases: prepare → cluster → refine → build result |
| `normalize.py` | Normalize / denormalize |
| `periods.py` | Unstack the flat series into period profiles |
| `clustering.py` | Adapt `ClusterConfig` onto the clustering kernels |
| `extremes.py` | Identify and inject extreme periods |
| `rescale.py` | Rescale representatives to preserve column means |
| `segmentation.py` | Merge adjacent timesteps into segments |
| `accuracy.py` | Reconstruct and score |
| `types.py` | The dataclass passed between each phase |

| `src/tsam/algorithms/` | Config-agnostic kernels |
|---|---|
| `clustering.py`, `k_medoids_exact.py`, `k_maxoids.py` | Assignment |
| `representations.py`, `duration_representation.py`, `concurrency.py`, `selection.py` | Representative selection |
| `segmentation.py` | Segment clustering |

Each function takes explicit inputs and returns explicit outputs — no shared mutable state. The phase boundaries follow what a caller can switch off: preparing and clustering always run, everything optional lives in `refine_representatives`, and `build_result` only expresses the outcome.

**The legacy API is gone.** `TimeSeriesAggregation`, `hyperparametertuning`, `LegacyAPIWarning` and `unstackToPeriods` are removed; `tsam.tuning` replaces `HyperTunedAggregations`.

**Weight handling consolidated**

v3 multiplied `weights` into the normalized series before unstacking, so every downstream step saw weighted data and divided it back out again — rescaling, reconstruction and accuracy each carried their own compensation. v4 applies them to the clustering candidates and removes them once, after the representative set is final.

The same steps are weighted either way (distance, medoid/maxoid selection, extreme detection, segment boundaries) and the same steps are not (rescale, denormalize, output, accuracy). Only the location of the division moved, so there is no observable change — see the breaking-changes section.

**Golden regression tests**

90 configurations × 4 datasets lock down numerical output. The baselines are generated by running **v3.4.2** through the same matrix, so they pin the reference implementation rather than whatever the current pipeline produces — v3.4.2 reproduces every pre-existing golden exactly, which is what makes that bridge trustworthy. Every configuration is additionally replayed through `ClusteringResult.apply()` to prove a stored clustering can reproduce its own run.

## Breaking changes

Each row below was checked by running the configuration on **v3.4.2** and on
this branch and comparing the output — not inferred from the code. Full
per-configuration guidance in
[`docs/migration/v3-to-v4.md`](docs/migration/v3-to-v4.md).

### Does this affect you?

Most of it needs nothing. These are the cases that do:

| If you… | What happens | Severity |
|---|---|---|
| set a per-column representation on **both** `ClusterConfig` and `SegmentConfig` | your v3 cluster representatives were built from the *segment* column assignment — the cluster one was ignored outright | **your v3 results were wrong, not just different** |
| rely on a `MinMaxMean` column name that does not exist, or list a column as both min and max | now a `ValueError` | **working v3 code stops running** |
| index result columns by position | columns follow the input order instead of being sorted alphabetically | reorder or `.sort_index(axis=1)` |
| transfer a clustering built with `scale_by_column_means` | the setting now survives the transfer; v3 silently dropped it | transferred results change |
| upgrade from **3.4.1 or earlier** using `distribution_minmax` / `maxoid` | the bounded water-fill applies (shipped in 3.4.2) | regenerate pinned references |
| pass a non-datetime index without `temporal_resolution` | defaults to hourly instead of raising | pass it explicitly |
| pin values that happened to land on a medoid tie | tie-breaking is deterministic now | it was never machine-independent before |
| use `TimeSeriesAggregation` or any internal module path | removed | move to `tsam.aggregate()` |

Weight handling and the water-fill were also reworked, but neither changes
results for anyone coming from 3.4.2 — see below.

### Verified to change behaviour

| Change | Who is affected | Evidence |
|---|---|---|
| Legacy `TimeSeriesAggregation` API removed | anyone not already on `tsam.aggregate()` | import fails |
| Package reorganised; v3 compat shims dropped | importers of internal modules | import fails |
| Output column order preserves input order | code indexing columns positionally | `GHI, Load, T, Wind` → `GHI, T, Wind, Load` |
| Cluster and segment representations resolved independently | `MinMaxMean` on **both** `ClusterConfig` and `SegmentConfig` | max abs **280.5** |
| `MinMaxMean` rejects duplicate/unknown column names | **code that ran silently in v3 now raises** | `ValueError` |
| Representative selection deterministic on ties | pinned values that landed on a tie | max abs **21.75**, 1 golden |

### Deprecations — old spelling still works

| Change | Behaviour |
|---|---|
| `cluster_weights` → `cluster_counts` | old name emits `FutureWarning` |
| `Distribution(scope="cluster")` → `"local"` | old value normalised, emits `FutureWarning` |

### Claimed as breaking, but measured identical to v3.4.2

Two changes the migration guide described as affecting results **do not differ
from v3.4.2 at all**, for two different reasons.

| Claim | Measured | Why |
|---|---|---|
| Weight decoupling changes medoid/maxoid representatives | identical — `1e-13` (medoid), `2e-13` (maxoid) | **never true of v4.** `representations()` receives the weighted candidates here exactly as in v3; unweighting happens later, in `refine_representatives`. Both versions select the representative in weighted space, so the choice cannot diverge. |
| Water-fill changes `distribution_minmax`, incl. with segmentation and without rescaling | identical — `1e-14` | **already released.** It shipped on the v3 line as #373 in **3.4.2**, where the changelog recorded it as "Conurrency and integral fix" under Bug Fixes with no sign that output moves. A change only from 3.4.1 or earlier. |

For `hierarchical_weighted` the agreement is total — same assignments, same
reconstruction, same selected medoid periods (`245, 78, 64, 17, 320, 201, 319,
263`).

`docs/migration/v3-to-v4.md` has been corrected for both.

## What the golden diff shows

`test/data/golden/` covers 90 configurations across 4 datasets, and the
baselines on `develop` were generated by v3.4.2 itself (#440). So the golden
section of this diff *is* the numerical delta between the two lines.

**It is one file.**

| File | Cells | Max abs |
|---|---|---|
| `segmentation_medoid/testdata.csv` | 572 / 35 040 | **21.75** |

That configuration has a segment of exactly two timesteps. Both members are
therefore equidistant, the medoid is mathematically undefined, and v3 resolved
it with a bare `argmin` decided by floating-point noise — the same input picked
a different representative depending on column order or BLAS build. v4 breaks
the tie deterministically, so the value moved once and stays put.

**Every other configuration is byte-identical between v3.4.2 and this branch.**

## One deletion worth a second look

`src/tsam/utils/k_medoids_contiguity.py` is gone, and with it the `networkx`
dependency.

It was never reachable from the public API — `contiguous` was and still is
`AgglomerativeClustering(linkage="ward", connectivity=…)`, and all seven
`contiguous` golden cases generated by v3.4.2 pass here unchanged. But it was
**not** unused code: `develop` carries `test/test_k_medoids_contiguity.py`,
which exercises both `k_medoids_contiguity` and `_contiguity_to_graph`. It is a
maintained implementation of a different and stronger algorithm — cutting-plane
adjacency-constrained k-medoids (Oehrlein & Hauner, 2017) — that looks like it
was meant to back `contiguous` and never got wired up.

Removing it is right for this PR: shipping a dependency for code no user can
reach is worse. But the capability, and its test, deserve an issue rather than
disappearing in a rewrite.

## Draft changelog

> Working draft for the v4 release notes — release-please will generate the final version from the commit history.

### ⚠ BREAKING CHANGES

* **api:** the legacy `TimeSeriesAggregation` class, `hyperparametertuning`, `LegacyAPIWarning` and `unstackToPeriods` are removed. `tsam.aggregate()` is the sole entry point and `tsam.tuning` replaces `HyperTunedAggregations`.
* **structure:** package reorganised into `pipeline/` and `algorithms/`; v3 backward-compat shims dropped.
* **output:** results preserve the input column order instead of sorting alphabetically.
* **representations:** `ClusterConfig.representation` and `SegmentConfig.representation` are resolved independently. v3 collapsed them into one per-column dict and let the segment one win, so setting `MinMaxMean` on both stages changes cluster representatives.
* **config:** `MinMaxMean` raises `ValueError` for a column listed in both `max_columns` and `min_columns`, and for a column that is not in the data. A misspelled name used to be a silent no-op.

> Weight handling was restructured but is not listed here: it produces identical
> output to v3.4.2 by construction. The bounded water-fill for the duration
> representation is likewise absent — it shipped in 3.4.2 as #373, not in v4.

### Features

* new pipeline architecture — stateless pure functions with typed milestones between phases (#176)
* `ClusteringResult` is serializable and transferable; `apply()` replays a stored clustering on new data
* `Distribution` and `MinMaxMean` representation objects for fine-grained per-column control
* concurrency strategies for the duration representation (`medoid`, `consensus`, `assignment`, `reference`)
* `KMedoids` forwards solver options (#365)
* lazy accuracy computation — `AggregationResult.accuracy` is computed on first access
* misconfigurations are reported: partial final periods warn that they are padded, and `representation` with `use_duration_curves` warns that it has no effect

### Bug Fixes

* **clustering:** duration-curve centers and their recorded indices describe the same periods; a stored clustering now replays to the run it came from
* **clustering:** period-sum features stay out of the duration-curve path, whose per-column reshape they shifted by one block
* **pipeline:** period-sum features are summed from the weighted profiles, so a column's sum carries the same weight as its timesteps
* **algorithms:** medoid and maxoid selection breaks ties deterministically — the choice no longer depends on column order or BLAS build
* **result:** `apply()` replays under the stored `ClusterConfig`, so `scale_by_column_means` survives a transfer
* **result:** `apply()` accepts a clustering built from a series with a padded partial last period
* **result:** transfers that cannot be reproduced exactly now warn, for both `extremes="replace"` and `append`/`new_cluster` with a computed representation
* **representations:** corrected maxoid variable name and documented its distance scope (#366)

### Documentation

* Diátaxis-structured documentation: tutorials, how-to guides, explanation and API reference
* architecture guide and pipeline diagram; ADR for the pipeline decomposition
* v3→v4 migration guide covering every behaviour change

## Test plan

- [x] 687 passed, 145 skipped, 5 xfailed (`pytest test/ -q`)
- [x] Golden regression over 90 configurations, baselines generated by v3.4.2
- [x] Every configuration replayed through `ClusteringResult.apply()`, with guards that a replay does not silently re-cluster
- [x] `test_weight_decoupling.py` verifies weights affect only clustering distance
- [x] `ruff check` / `ruff format --check` and `mkdocs build --strict` clean


FBumann:
@julian-belina I merged my rewrite into here.
Im open for adjustments and improvements, both to the code structure and to the api.
Ill track remaining work/issues in a milestone https://github.com/FZJ-IEK3-VSA/tsam/milestone/2

FBumann:
@julian-belina Any updates?
#235 #255 #256 #257 wait to merged into here. But only after your initial review!

julian-belina:
@FBumann, it's on my to-do list, but I'm super busy unfortunately. I hope I can get to it by the end of the week or next week. It's definitely not forgotten!

FBumann:
@julian-belina No worries, just checking back. I wont have time either the next few weeks to tackle those issues.

l-kotzur:
This looks really promising!

FBumann:
@l-kotzur great to hear!

Due to the existing sophisticated testing im also very confident it preserves behaviour entirely!

FBumann:
@julian-belina I ported the latest additions to tsam over to the new pipeline.

FBumann:
@julian-belina The merging of #373 produced a lot of churn.
I hope that this didnt corrupt anything!
Resolving the conflicts was quite a hustle.

FBumann:
@julian-belina After we merged #422 Im happy

FBumann:
After the workflows finish we can merge imo! @julian-belina

src/tsam/pipeline/orchestrator.py
```diff
@@ -0,0 +1,679 @@
...
+
+
+def _build_representation_dict(
+    columns: pd.Index,
+    cluster_representation: str | Distribution | MinMaxMean | None,
+) -> dict[str, str]:
+    """Build the representation dict (mean/min/max per column) from config."""
+    from tsam.config import MinMaxMean
+
+    representation_dict: dict[str, str] = dict.fromkeys(columns, "mean")
```
julian-belina, line 191: Currently, if a column is should preserve "min" and "max" values the max value in the representation_dict is just overwritten. Its a misconfiguration but should be caught early on. Also its not compatible with distribution and medoid

src/tsam/pipeline/periods.py
```diff
@@ -0,0 +1,142 @@
...
+            input).
+
+    Note:
+        `add_period_sum_features` optionally appends per-period column sums as
+        extra clustering features.
+    """
+    unstacked = normalized_ts.copy()
+
+    # Extend to integer multiple of period length
+    if len(normalized_ts) % n_timesteps_per_period == 0:
```
julian-belina: Is there a case in which the time series is not an integer multiple of the period length at this point in the pipeline? Isn't this caught in line 172 of the aggregate function? Raising an error instead of inventing data would be better, I think. Or is this behavior intended?

src/tsam/pipeline/periods.py
```diff
@@ -0,0 +1,142 @@
...
+        n_columns=n_columns,
+        n_periods=n_periods,
+    )
+
+
+def add_period_sum_features(
+    profiles_df: pd.DataFrame,
+    candidates: np.ndarray,
+) -> tuple[np.ndarray, int]:
+    """Append each period's per-column sum as extra clustering features.
```
julian-belina, line 123: I think it would be helpful if the docstring stated that the function calculates the sum of the normalized values. This becomes clear when investigating the whole pipeline, but it can be confusing when looking at the function individually.

julian-belina: It's also strange that the n_extra variable is calculated but never used.

src/tsam/pipeline/orchestrator.py
```diff
@@ -0,0 +1,679 @@
...
+
+    Note:
+        The candidates clustered here come from ``prepare_data``; the resulting
+        representatives are consumed by ``format_and_reconstruct``.
+    """
+    cluster = cfg.cluster
+    cluster_representation = cluster.get_representation()
+    candidates = prepared.candidates
+    period_profiles = prepared.period_profiles
+    reference_attribute_idx = _resolve_reference_attribute_idx(
```
julian-belina, line 376: This function is a bit off. Wouldn't it make more sense to retrieve the column list from the PreparedData object?

src/tsam/pipeline/orchestrator.py
```diff
@@ -0,0 +1,679 @@
...
+
+def _resolve_reference_attribute_idx(
+    columns: list,
+    cluster_representation: str | Distribution | MinMaxMean | None,
+) -> int | None:
+    """Map a ``Distribution.reference_attribute`` name to its attribute index.
+
+    ``columns`` must be in the same order as the attribute blocks of the
+    clustering candidates (i.e. the representation dict order), so the returned
+    index lines up with the duration representation.
```
julian-belina: Adding a description of the return type would help people understand what this function is supposed to do more quickly.

src/tsam/pipeline/clustering.py
```diff
@@ -0,0 +1,242 @@
...
+        return (
+            candidates[predef.cluster_center_indices],
+            list(predef.cluster_center_indices),
+            predef.cluster_order,
+        )
+    else:
+        centers, computed_indices = representations(
+            candidates,
+            predef.cluster_order,  # type: ignore[arg-type]
+            default="medoid",
```
julian-belina: I don't think it's good practice to set the default methods in multiple locations. This should be inferred from the configuration object.

src/tsam/algorithms/representations.py
```diff
@@ -0,0 +1,224 @@
...
+from tsam.algorithms.duration_representation import duration_representation
+from tsam.config import Distribution, MinMaxMean
+
+
+def representations(
+    candidates: np.ndarray,
+    cluster_order: np.ndarray,
+    default: str,
+    representation_method: str | Distribution | MinMaxMean | None = None,
+    representation_dict: dict[str, str] | None = None,
```
julian-belina, line 16: The representation_dict is ignored for the distribution, medoid and maxoid options. I also think it's not trivial to implement respecting the representation_dict for these methods. It's probably a good idea to flag it as a misconfiguration when both are selected early on.

src/tsam/pipeline/clustering.py
```diff
@@ -0,0 +1,242 @@
...
+
+from tsam.algorithms.clustering import cluster_and_represent
+from tsam.algorithms.representations import representations
+
+if TYPE_CHECKING:
+    from tsam.config import ClusterConfig, Distribution, MinMaxMean
+    from tsam.pipeline.types import PredefParams
+
+
+def cluster_periods(
```
julian-belina, line 29: This method almost a pure wrapper without own contribution. Maybe we can get rid of it and call cluster_and_represent directly?

src/tsam/pipeline/clustering.py
```diff
@@ -0,0 +1,242 @@
...
+    # Sort each period's timesteps descending for all columns.
+    # Use candidates (already weighted) so that clustering distance respects
+    # column weights — matching v3 behaviour.
+    n_periods, n_total = candidates.shape
+    n_timesteps = n_total // n_columns
+
+    values_3d = candidates.copy().reshape(n_periods, n_columns, n_timesteps)
+    sorted_values = (-np.sort(-values_3d, axis=2, kind="stable")).reshape(n_periods, -1)
+
+    _, center_indices, cluster_order = cluster_and_represent(
```
julian-belina: Variable names that are a bit more expressive than "_", would be nice

src/tsam/pipeline/clustering.py
```diff
@@ -0,0 +1,242 @@
...
+        solver=cluster.solver,
+        cluster_method=cluster.method,
+        representation_method=cluster.get_representation(),
+        representation_dict=representation_dict,
+        n_timesteps_per_period=n_timesteps_per_period,
+        reference_attribute_idx=reference_attribute_idx,
+    )
+
+    # Pick medoid from unsorted candidates (already weighted).
+    cluster_centers = []
```
julian-belina, line 210: Why are medoids always picked instead of the configured representation method?

src/tsam/algorithms/clustering.py
```diff
@@ -0,0 +1,133 @@
...
+    Two steps: :func:`assign_clusters` assigns each period to a cluster, then
+    :func:`~tsam.algorithms.representations.representations` derives the cluster
+    representatives (using the per-method default unless ``representation_method``
+    overrides it).
+    """
+    cluster_order = assign_clusters(
+        candidates, n_clusters, cluster_method, n_iter=n_iter, solver=solver
+    )
+
+    # Representatives may be drawn from a separate candidate set (e.g. unweighted).
```
julian-belina:
```suggestion
    # Representatives may be drawn from a separate candidate set (e.g. unweighted or with included period sums).
```

src/tsam/pipeline/extremes.py
```diff
@@ -0,0 +1,210 @@
...
+def _detect_extreme(
+    profiles_df: pd.DataFrame,
+    column: str | tuple,
+    series: pd.Series,
+    suffix: str,
+    extreme_period_no: list,
+    cc_list: list,
+    extreme_periods: dict,
+) -> None:
+    """Detect a single extreme period and add it to extreme_periods if unique."""
```
julian-belina: More expressive docstrings would help to make this function more understandable.

src/tsam/pipeline/extremes.py
```diff
@@ -0,0 +1,210 @@
...
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    import numpy as np
+    import pandas as pd
+
+    from tsam.config import ExtremeConfig
+
+
+def _append_col_with(column: str | tuple, append_with: str = " max.") -> str | tuple:
```
julian-belina: More expressive docstrings would help to make this function more understandable.

src/tsam/pipeline/extremes.py
```diff
@@ -0,0 +1,210 @@
...
+        `cluster_periods` produces the clusters this stage augments, and
+        `rescale_representatives` skips extreme clusters when correcting means.
+    """
+    columns = profiles_df.columns.get_level_values(0).unique().tolist()
+    extreme_periods: dict = {}
+    extreme_period_no: list = []
+
+    cc_list = [center.tolist() for center in cluster_centers]
+
+    # Detect extreme periods for each column
```
julian-belina, line 112:
```suggestion
    # Detect extreme periods for each column
    # (lists_of_column_names_to_consider, extreme_variant, column_suffix)
```

src/tsam/pipeline/extremes.py
```diff
@@ -0,0 +1,210 @@
...
+    new_cluster_centers: list = []
+    new_cluster_order = list(cluster_order)
+    extreme_cluster_idx: list[int] = []
+
+    if extremes.method == "append":
+        for cluster_center in cluster_centers:
+            new_cluster_centers.append(cluster_center)
+        for i, period_type in enumerate(extreme_periods):
+            extreme_cluster_idx.append(len(new_cluster_centers))
+            new_cluster_centers.append(extreme_periods[period_type]["profile"])
```
julian-belina: Not chaining the operations inline makes them much easier to read.

src/tsam/pipeline/extremes.py
```diff
@@ -0,0 +1,210 @@
...
+    elif extremes.method == "new_cluster":
+        for cluster_center in cluster_centers:
+            new_cluster_centers.append(cluster_center)
+        for i, period_type in enumerate(extreme_periods):
+            extreme_cluster_idx.append(len(new_cluster_centers))
+            new_cluster_centers.append(extreme_periods[period_type]["profile"])
+            extreme_periods[period_type]["new_cluster_no"] = i + len(cluster_centers)
+
+        # Build set of extreme period step numbers for quick lookup
+        extreme_step_nos = {extreme_periods[pt]["step_no"] for pt in extreme_periods}
```
julian-belina: Same here less inline chaining would it make much easier to read

src/tsam/pipeline/orchestrator.py
```diff
@@ -0,0 +1,679 @@
...
+    normalized_typical_periods = _representatives_to_dataframe(
+        clustered.cluster_periods_list, period_profiles.column_index
+    )
+
+    # Segmentation if configured
+    segmented_df = None
+    predicted_segmented_df = None
+    segment_center_indices = None
+
+    if cfg.segments is not None:
```
julian-belina, line 556: I think the segmentation in format_and_reconstruct() is a bit unintuitive. Maybe having a firth step between cluster_and_postprocess() and format_and_reconstruct() that includes extreme periods, rescaling, and segmentation would be a better call since these are all optional

src/tsam/pipeline/orchestrator.py
```diff
@@ -0,0 +1,679 @@
...
+    return FormattedOutput(
+        typical_periods=typical_periods,
+        reconstructed_data=reconstructed_data,
+        normalized_predicted=normalized_predicted,
+        segmented_df=segmented_df,
+        segment_center_indices=segment_center_indices,
+    )
+
+
+def assemble_result(
```
julian-belina: Shouldnt assemble_result() be merged with format_and_reconstruct() ?

julian-belina (commented):
Hey @FBumann,

Sorry for the delayed review and thank you for bearing with me. This is really great work and I think the new architecture is much better than the V3 architecture! However, there are some things that should be addressed before we can release V4, which I addressed in the comments on this review.

To prevent the need to resolve any more merge conflicts, I am open to merging the V4 branch into the develop branch. First, though, I would like to create a deprecated V3 branch in case I need to make any comparisons in the next month. Once these comments are adressed we can do the first V4 Release.

src/tsam/pipeline/clustering.py
FBumann: _ is the convention for sth that's not used anymore. As we are getting a tulle returned, it's the best way here. Any proper name would be misdirection.

FBumann:
Hi @julian-belina,

I'll address your review.

I think It can be divided into style opinions, documentation and stuff I need to look into.

## About the styling: 

I argue inlining is denser and better than creating intermediate variables just to discard them right away. But that's an opinion. I'm open for suggestions, but I wouldn't rewrite it.

_ is the convention for unpacking if the unpacked variable is not needed.



## About documentation

Could you supply suggestions for those? Or should I just do it?

## About the others

I'll address them with PRs or in the review comments

julian-belina:
@FBumann: The styling and documentation go hand in hand for some functions. Short, non-expressive variable names crammed into a line with missing docstrings are very hard to review. 

I was not aware of the convention, but we can stick to it. However, there is another comment related to the same section that says it's unclear why the variable was dropped.

FBumann:
**Reminder before the v4 release:** of the 21 commits in #434–#439, only one carries a `BREAKING CHANGE:` footer, so release-please will file the other six result-changing ones under *Bug Fixes* instead of *⚠ BREAKING CHANGES* — including `MinMaxMean` typos now raising `ValueError`, which breaks code that ran silently in v3.

All six are documented in `docs/migration/v3-to-v4.md` and the major bump is unaffected; only the generated changelog is. Details in #439.

FBumann:
@julian-belina I merged everything. Im now going to rewrite the PR descritpion. Then we can start the release workflow. But we need to write a Cahngelog entry manually in the release please PR then! (just edit the PR description, its gonna be the changelog/release notes!

We agreed on squashing this into develop!

FBumann:
@julian-belina Im gonna do one last check:
Im going to open a that diffs the goldens of this branch with the goldens of 3.4.2, backporting all of them onto develop.

FBumann:
@julian #440 showed that only one breaking change is the stable sort in segmentation!
The other got backported to 3.4.2 in "Conurrency and integral fix (#373)"

This is now ready to be squashed finally!

FBumann:
CLAUDE:


**Correction to my earlier comment.** I said six commits needed `BREAKING CHANGE:` footers. Having since measured each configuration on v3.4.2 against this branch, it is **two**, and both remaining ones are already accounted for:

| Originally listed | Status |
|---|---|
| `MinMaxMean` raising `ValueError` | real — footer added in `5d2dedf` |
| cluster/segment representation split | real — footer added in `21b7831` |
| period-sum features weighted | **not breaking** — restores v3 behaviour |
| period sums out of duration-curve clustering | **not breaking** — restores v3 behaviour |
| duration-curve center indices | metadata and transfers only; `aggregate()` output is unchanged, and it sits inside `5d2dedf` |
| deterministic tie-breaking | real, no footer — `5a8d49b` |

Plus one I had not listed: `fe80a46`, `apply()` replaying the stored `ClusterConfig`, which changes transferred results when `scale_by_column_means` is set.

So the outstanding gap is `5a8d49b` and `fe80a46`. Everything else either carries a footer or does not belong on the list.

The wider point behind the correction: two claims in `docs/migration/v3-to-v4.md` did not survive measurement either — weight decoupling produces identical output *by construction* (both versions select the representative in weighted space), and the water-fill shipped in **3.4.2** as #373 rather than in v4. Both sections are corrected on this branch.

Across all 90 golden configurations, v3.4.2 and this branch differ in **one file**: `segmentation_medoid`, a segment of two equidistant timesteps where the medoid is undefined.

julian-belina (approved):

PR #267, opened 2026-04-01
feat: preserve DatetimeIndex through aggregate/disaggregate round-trip

### LINKED ISSUE #266: Preserve original DatetimeIndex through aggregate/disaggregate round-trip
## Problem

`tsam.aggregate()` receives a DataFrame with a `DatetimeIndex`, but `ClusteringResult.disaggregate()` returns a DataFrame with an integer index `(0, 1, 2, ...)`. The original time index is lost.

This means downstream code must store the time index separately and reassign it after disaggregation (which is what `tsam_xarray` currently does via `ClusteringResult.time_coords`).

## Proposed solution

1. **Store the original index** in `ClusteringResult` during `aggregate()` — e.g. as a `time_index` attribute.
2. **Restore it** in `disaggregate()` — just assign it to the output DataFrame's index.
3. **Serialize it** in `to_dict()`/`from_dict()` — for regular indices, store compactly as `{start, periods, freq}` (~100 bytes) instead of a full list (~200 KB for 8760h). Fall back to a list for irregular indices.

## Why

- `disaggregate()` output becomes directly usable with time-aware operations (plotting, resampling, joining)
- Serialization round-trips preserve the time index
- Downstream wrappers (e.g. tsam_xarray) can simplify by dropping their own time index bookkeeping

## Scope

- Additive change — no existing behavior breaks
- `time_index` can default to `None` for backward compatibility with older serialized files
- Only affects `ClusteringResult`: new attribute + small changes to `disaggregate()`, `to_dict()`, `from_dict()`

## Compact serialization sketch

```python
# to_dict
freq = pd.infer_freq(time_index)
if freq is not None:
    data["time_index"] = {"start": time_index[0].isoformat(), "periods": len(time_index), "freq": freq}
else:
    data["time_index"] = [t.isoformat() for t in time_index]

# from_dict
raw = data.get("time_index")
if isinstance(raw, dict):
    time_index = pd.date_range(raw["start"], periods=raw["periods"], freq=raw["freq"])
elif isinstance(raw, list):
    time_index = pd.DatetimeIndex(raw)
else:
    time_index = None
```

FBumann:
## Summary
- Adds `time_index` field to `ClusteringResult` that stores the original `DatetimeIndex` from `aggregate()` input data
- `disaggregate()` restores the time index on the output DataFrame when available
- `to_dict()`/`from_dict()` serialize the index compactly (`{start, periods, freq}` for regular indices, ISO list fallback for irregular ones)
- Fully backward-compatible: `time_index` defaults to `None`, old serialized files work unchanged

Closes https://github.com/FZJ-IEK3-VSA/tsam/issues/266

## Test plan
- [x] New test: `test_datetime_index_restored` — disaggregate returns DatetimeIndex matching original
- [x] New test: `test_integer_index_when_no_time_index` — RangeIndex when no time_index stored
- [x] New test: `test_io_roundtrip_preserves_time_index` — JSON round-trip preserves DatetimeIndex
- [x] New test: `test_io_roundtrip_no_time_index` — backward compat with old files
- [x] All 477 existing tests pass

### DIFF
--- a/test/test_disaggregate.py
+++ b/test/test_disaggregate.py
@@ -225,6 +225,42 @@ class TestClusteringResultDisaggregate:
         np.testing.assert_array_equal(original.values[mask], restored.values[mask])
 
 
+class TestTimeIndexSerialization:
+    """Unit tests for _time_index_to_dict / _time_index_from_dict helpers."""
+
+    def test_regular_index_compact(self):
+        from tsam.config import _time_index_to_dict
+
+        idx = pd.date_range("2025-01-01", periods=8760, freq="h")
+        d = _time_index_to_dict(idx)
+        assert isinstance(d, dict)
+        assert set(d) == {"start", "periods", "freq"}
+        assert d["periods"] == 8760
+
+    def test_regular_index_roundtrip(self):
+        from tsam.config import _time_index_from_dict, _time_index_to_dict
+
+        idx = pd.date_range("2025-01-01", periods=8760, freq="h")
+        restored = _time_index_from_dict(_time_index_to_dict(idx))
+        pd.testing.assert_index_equal(idx, restored)
+
+    def test_irregular_index_fallback(self):
+        from tsam.config import _time_index_from_dict, _time_index_to_dict
+
+        idx = pd.DatetimeIndex(["2025-01-01", "2025-01-03", "2025-01-07"])
+        d = _time_index_to_dict(idx)
+        assert isinstance(d, list)
+        restored = _time_index_from_dict(d)
+        pd.testing.assert_index_equal(idx, restored)
+
+    def test_old_list_format_still_loads(self):
+        from tsam.config import _time_index_from_dict
+
+        raw = ["2025-01-01T00:00:00", "2025-01-01T01:00:00"]
+        restored = _time_index_from_dict(raw)
+        assert len(restored) == 2
+
+
 class TestDisaggregateEdgeCases:
     """Edge cases and robustness tests."""
 
--- a/test/test_disaggregate.py
+++ b/test/test_disaggregate.py
@@ -105,8 +105,20 @@ class TestAggregationResultDisaggregate:
 class TestClusteringResultDisaggregate:
     """Tests for ClusteringResult.disaggregate."""
 
-    def test_integer_index(self, result):
+    def test_datetime_index_restored(self, result):
         expanded = result.clustering.disaggregate(result.cluster_representatives)
+        if result.clustering.time_index is not None:
+            assert isinstance(expanded.index, pd.DatetimeIndex)
+            assert expanded.index.equals(result.clustering.time_index)
+        else:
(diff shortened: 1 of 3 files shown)


test/test_disaggregate.py
```diff
@@ -187,6 +225,42 @@ def test_io_roundtrip_segmented(self, result_segmented, tmp_path):
...
+    def test_regular_index_compact(self):
+        from tsam.config import _time_index_to_dict
+
+        idx = pd.date_range("2025-01-01", periods=8760, freq="h")
+        d = _time_index_to_dict(idx)
+        assert isinstance(d, dict)
+        assert set(d) == {"start", "periods", "freq"}
+        assert d["periods"] == 8760
+
+    def test_regular_index_roundtrip(self):
```
julian-belina, line 240: It would be helpful to split these tests into serialization and deserialization tests. This would allow us to inspect the persisted file on the hard drive. Is the frequency supposed to be stored in a CSV file or a JSON file in addition to the CSV?

FBumann: in a json. As part of the `to_json` and `to_dict` methods

julian-belina: But is the frequency supposed to be stored explicitly on the hard drive? I thought so because you mentioned the byte size of the indices.

FBumann: If you save it to json, it will be stored in a file.

FBumann: Its a from and to dict serialization.
The json step is on top. The optimization here reduces the size of the object or file by not having to stire each timestep as a ISO-Date string, but just store the start, count and freq. This is sufficient if its equidistant

julian-belina: I don't think I understand the point of creating and reading the index by frequency and save it as a dictionary. Why not just save it as it is?

julian-belina: Is the goal to save memory during TSAM execution?

FBumann: The reason for the PR is to make `ClusteringResult` self contained. It  now can `disaggregate()` data back to its original form (with the time index).
We support **IO** for `ClusteringResult`, both `json` and `dict`. So we need to store the time index.
Such a DatetimeIndex is quite large, as it needs t store a str like '2020-01-01 14:00:00...' for every timestep.
So i thought ill store it more efficiently. THis is a pure size optimization for the `ClusteringResult` class, which does not hold the data, but how the clustering maps from timesteps to periods/clusters/representatives.

No relevant memory savings.
It will only matter if you save 10 or more of such objects as a dict/json somewhere in another object or an actual file.

FBumann: > I don't think I understand the point of creating and reading the index by frequency and save it as a dictionary. Why not just save it as it is?

What do you mean with save as is? As a list of strings? Or using pandas `to_dict()`(viable, but large)?

julian-belina (approved):

PR #277, opened 2026-04-16
fix: KeyError in accuracyIndicators() when weightDict is a column subset

### LINKED ISSUE #276: ClusteringResult.apply() fails with KeyError when data has columns not in weights
## Description

`ClusteringResult.apply(data)` raises a `KeyError` when the new `data` DataFrame has columns that were not present in the original clustering's `weights` dict.

This happens because `apply()` passes `weights=self.weights` to `TimeSeriesAggregation`, and then `accuracyIndicators()` (line 1447 of `timeseriesaggregation.py`) does `self.weightDict[column]` for **every** column in the new data — but `weightDict` only has entries from the original clustering.

## Minimal reproducible example

```python
import pandas as pd
import numpy as np
import tsam

np.random.seed(42)
idx = pd.date_range('2024-01-01', periods=168, freq='h')
df_subset = pd.DataFrame({'wind': np.sin(np.linspace(0, 14*np.pi, 168)) + 2}, index=idx)
df_full = df_subset.copy()
df_full['solar'] = np.cos(np.linspace(0, 10*np.pi, 168)) + 1.5

# Cluster on subset with explicit weights
result = tsam.aggregate(
    df_subset, n_clusters=2, period_duration=24, temporal_resolution=1,
    weights={'wind': 1.0},
)

# Apply clustering to full data -> KeyError: 'solar'
result.clustering.apply(df_full)
```

## Traceback

```
File "tsam/config.py", line 1066, in apply
    accuracy_df = agg.accuracyIndicators()
File "tsam/timeseriesaggregation.py", line 1447, in accuracyIndicators
    origTS = self.normalizedTimeSeries[column] / self.weightDict[column]
KeyError: 'solar'
```

## Expected behavior

`apply()` should handle columns not in `weights` — either by defaulting missing columns to weight `1.0`, or by computing accuracy only for columns that have weights.

## Environment

- tsam 3.3.0
- Python 3.11

FBumann:
## Summary
- `accuracyIndicators()` raised `KeyError` when `weightDict` didn't contain all columns in the data (e.g. when `apply()` is used with extra columns)
- Use `.get(column, 1)` instead of `[column]` so unweighted columns default to weight 1
- Added regression test

Closes #276

## Test plan
- [x] Reproducer from issue no longer raises `KeyError`
- [x] New test `test_apply_with_weights_and_extra_columns` passes
- [x] All existing tests pass (481 passed)

### DIFF
--- a/src/tsam/timeseriesaggregation.py
+++ b/src/tsam/timeseriesaggregation.py
@@ -1444,7 +1444,9 @@ class TimeSeriesAggregation:
 
         for column in self.normalizedTimeSeries.columns:
             if self.weightDict:
-                origTS = self.normalizedTimeSeries[column] / self.weightDict[column]
+                origTS = self.normalizedTimeSeries[column] / self.weightDict.get(
+                    column, 1
+                )
             else:
                 origTS = self.normalizedTimeSeries[column]
             predTS = self.normalizedPredictedData[column]
--- a/test/test_clustering_e2e.py
+++ b/test/test_clustering_e2e.py
@@ -438,6 +438,19 @@ class TestClusteringTransfer:
             input_data.columns
         )
 
+    def test_apply_with_weights_and_extra_columns(self, input_data):
+        """Test applying clustering when data has columns not in weights (GH-276)."""
+        # Cluster on a single column with explicit weights
+        wind_only = input_data[["Wind"]]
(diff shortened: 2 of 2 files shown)


julian-belina (approved):

julian-belina:
good catch sorry for the long review delay...

PR #282, opened 2026-04-21
fix integral and min max preservation

l-kotzur:
## Description

In rare cases tsam violates the original lower or upper bounds of the time series leading to a UserWarning: At least one maximal value of the aggregated time series exceeds the maximal value the input time series for: ...

## Motivation and Context

This fixes the scaling and the min max distribution representation.

## Type of Change

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [ ] My code follows the project's code style (run `ruff check` and `ruff format`)
- [X] I have added tests that prove my fix/feature works
- [ ] All new and existing tests pass (`pytest test/`)
- [ ] I have updated the documentation if needed


FBumann:
@l-kotzur I really like this. But i think we need to regenerate the golde test files for this to pass ci.
Also, it seems like this could hurt performance due to the 20 iterations.
Furthermore, would you mind adding this after #234 is merged? It seems to me that such changes will be much more easy to implement there, and each change while #234 sits waiting leads to extra work there.

But i fully understand if you want to merge this before of course :)

julian-belina:
@l-kotzur, thank you very much for looking into this. I hope I can review and merge it by the end of the week.

l-kotzur:
Thanks, I have no hurries. Do the #234 first. Then we can remove the golde test and discuss performance.

FBumann:
@l-kotzur just to make sure:
The golden regression tests are there to stay! They wont get removed. 
In cases like this where a bug is fixed that leads to different results, we need to simply decide:
1. Fix the bug by updating the golden regression files and merge the bugfix. THis will need proper release notes, as results of users will differ
2. Defer the bugfix, add a config option to opt in or do sth else to make it non-breaking

Option 1 should be prefered if the differences are as small as in this case, and if its an actual bug...

Option 2 adds complexity and should only be chosen for good reasons

FBumann:
@l-kotzur I think this is superseeded by #373, right? If so, id close this and discuss the rest there. Feel free to reopen

PR #289, opened 2026-05-13
ci: run pre-commit hooks in CI

FBumann:
## Summary
- Replace the standalone `ruff` lint job with `pre-commit/action@v3.0.1` in both CI workflows, so all hooks from `.pre-commit-config.yaml` (ruff, ruff-format, mypy, nbstripout, etc.) run on PRs.
- First commit applies the existing trailing-whitespace / EOF auto-fixes (README, 2 SVGs, CSS).

Check-only — fails CI if a hook modifies a file; does not push fix commits back. Context: #287.

## Renovate updates (added in follow-up commits)

- **`pre-commit` manager** enabled, so Renovate bumps hook `rev:` pins in `.pre-commit-config.yaml`. Grouped into one PR per run.
- **Conda-forge gating for shared deps**: a custom regex manager in `.github/renovate.json` resolves the 21 deps shared between `pyproject.toml` and `environment.yml` through conda-forge, and a `packageRule` disables `pep621` for the same deps. Net effect: a release on PyPI alone no longer opens a PR — Renovate waits until conda-forge has it, which prevents the situation in #287 where mypy 2.0 was proposed before conda-forge had it.
- **Maintenance contract documented** in `pyproject.toml`, `environment.yml`, and `.github/renovate.json`: adding a new dual-channel dep requires updating all three.
- **pre-commit.ci guard**: a `ci: autoupdate_schedule: never` block in `.pre-commit-config.yaml` so that if pre-commit.ci is ever enabled, it won't race with Renovate's pre-commit manager.
- Config schema validated locally with `npx renovate-config-validator --strict`.

PyPI-only deps (`pytest-benchmark`, `mkdocstrings[python]`, `pandas-stubs`, `notebook`) remain on the `pep621` manager.

## ⚠️ Required after merge

- If branch protection requires the `Lint` status check on `develop`/`master`, update it to require **`Pre-commit`** instead.
- The first Renovate run after merge may produce a small wave of one-off PRs as deps are re-attributed from `pep621` to the new custom regex manager. These are safe to close if duplicates of existing open Renovate PRs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #290, opened 2026-05-13
chore: migrate from mypy to ty for static type checking

FBumann:
## Summary
Stacked on top of #289. Swaps mypy for [ty](https://github.com/astral-sh/ty) (Astral's new type checker) and fixes all 28 diagnostics it surfaces (mypy reported 2).

- Drop mypy pre-commit hook (ty needs the real project venv, not pre-commit's isolated env).
- Add a dedicated `type-check` CI job in both workflows that installs `.[develop]` and runs `ty check src/tsam/`.
- Replace `mypy` with `ty<=0.0.35` in develop extras; swap `[tool.mypy]` for `[tool.ty.*]` in `pyproject.toml`.
- Fix 28 diagnostics across 9 files (callable narrowing, None-safety, pandas-stubs workarounds). All 482 tests pass.

Context: #287 discussion.

## ⚠️ Required after merge
If branch protection requires a `Type check` status, add it. Update any other status-check requirements for develop/master accordingly.

## Notes
- Rebase onto `develop` once #289 merges and re-target base.
- ty is alpha (currently 0.0.35); pinned with upper bound so future changes don't surprise CI.

### DIFF
--- a/.github/workflows/ci-develop.yaml
+++ b/.github/workflows/ci-develop.yaml
@@ -112,10 +112,26 @@ jobs:
       - name: Build MkDocs
         run: mkdocs build --strict
 
+  type-check:
+    name: Type check (ty)
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v6
+      - uses: astral-sh/setup-uv@v8.1.0
+        with:
+          enable-cache: true
+      - uses: actions/setup-python@v6
+        with:
+          python-version: "3.12"
+      - name: Install
+        run: uv pip install --system ".[develop]"
+      - name: Run ty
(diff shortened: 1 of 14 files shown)


FBumann:
@julian-belina Here is a sketch on the migration to ty. Works, surfaces issues, but im not sure if its worth it...

julian-belina:
@FBumann: You're right; it doesn't look quite finished yet. Maybe give them some more time to develop ty. However, I still think it would be a good idea to conduct pre-commit checks in the CI.

FBumann:
@julian-belina I would also argue it would delay #234 
aybe we can check after that merges...?

PR #323, opened 2026-05-28
docs: 315 add architecture description

julian-belina:
This pull request addresses part of the following issue: https://github.com/FZJ-IEK3-VSA/tsam/issues/315 by adding an architecture description.
I used [d2](https://d2lang.com/) to render the diagrams and welcome feedback on them. The source code for the diagrams is included alongside them.

- To improve local execution speed for the documentation build, I implemented parallel notebook execution and caching.
- Due to the caching introduced by the local execution, the notebooks now contain cached artifacts.
- I also added an option to enlarge diagrams using diagram-zoom-init.js.
- I also added the contributors guide to the Readthedocs documentation 

You can inspect the new rendered documentation here https://tsam.readthedocs.io/en/315-add-architecture-description/

PR #325, opened 2026-05-28
refactor: consolidate pipeline orchestration into run_pipeline() (#324)

### LINKED ISSUE #324: refactor: consolidate pipeline business logic into run_pipeline()
### Purpose of the improvement

# Summary
The architecture description added in #323 makes it clear that the pipeline consists of a well-defined, sequential series of stages:
normalize → periods → clustering → extremes → rescale → segment → accuracy
However, the business logic governing how and when each stage is executed is currently scattered across multiple locations in the codebase rather than being expressed in one authoritative place. This makes it harder to understand, test, and modify the pipeline as a whole.

The individual stage functions (normalize, periods, clustering, etc.) should remain as pure, self-contained units that only perform their own transformation — with no knowledge of the surrounding pipeline.
# Motivation

Readability: A developer new to the codebase should be able to read run_pipeline() and immediately understand the full execution flow without tracing calls across multiple files.
Testability: A single orchestration function makes it straightforward to write unit tests for the pipeline flow independently of the individual stage implementations.
Consistency with the documented architecture: The architecture diagram in #323 already depicts a clear "Pipeline orchestrator" node. The code should reflect this.

### Proposal

Move all pipeline business logic into a single run_pipeline() function in pipeline/orchestrator.py (or equivalent). This function should be the single source of truth for:

the order in which stages are executed
any conditional branching (e.g. whether the extremes or segment stage is run)
how PipelineConfig is consumed and how PipelineResult is assembled
error handling and validation between stages

FBumann:
## Summary

Closes #324.

Consolidates the pipeline business logic into a single `run_pipeline()` function as proposed in the issue, plus three follow-on readability cleanups in the same module. Stage modules (`normalize`, `periods`, `clustering`, `extremes`, `rescale`, `segment`, `accuracy`) were already pure transforms — they are not touched.

Four commits, one concern each:

- **f8bb2d1** — Inline the four phase helpers (`_prepare_data`, `_cluster_and_postprocess`, `_format_and_reconstruct`, `_assemble_result`) into `run_pipeline()`. Delete the three internal milestone dataclasses (`PreparedData`, `ClusteringOutput`, `FormattedOutput`) that only existed to give those helpers typed return values. Every stage call, conditional branch, and PipelineConfig → PipelineResult step is now visible in one function.
- **b9229ba** — Hoist `from tsam.config import ClusteringResult, MinMaxMean` out of function bodies and into module-level imports. The lazy imports were workarounds for a circular dep that no longer exists (`tsam.config` doesn't eagerly import `tsam.pipeline`).
- **a8dd6b2** — Replace three weight-state variables and four touch points with a small `_Weighter` helper that owns the vector/dict forms and exposes apply/remove methods for each data shape (candidates ndarray, cluster centers, MultiIndex profiles DataFrame, flat typical-periods DataFrame). The orchestrator no longer names a weight vector or a tile array.
- **c325287** — Rephrase banner comments from `Step 1 … Step 14` to descriptive names. The numbering was a hold-over from the 16-step framing; with the four-phase split gone, contributors had a redundant index to keep in sync with the docs.

Diff: **+195 / -288** (net 93 lines smaller) on top of `develop-v4`, despite adding `_Weighter`.

## Test plan

- [x] `pytest test/ -q` — 559 passed, 112 skipped, 1 xpassed (unchanged from baseline at every commit)
- [x] `pre-commit run --all-files` — ruff, ruff-format, mypy all pass
- [ ] CI run on this PR
- [ ] Review of `run_pipeline()` end to end — does the one-function flow actually read better than the four-phase split?

## Notes

- The free function `_build_weight_vector` is retained at module scope because `test/test_weight_decoupling.py::TestBuildWeightVector` exercises it as a unit. `_Weighter` calls into it.
- Architecture docs (`docs/background/architecture/components.md`, `pipeline_guide.md`) still describe the four-phase split — they were merged in #323 and will need a follow-up edit once that PR lands and this rebases onto its merge commit. Suggesting that as a separate doc-only PR rather than mixing in here.
- One follow-on item I would still recommend, not included here: `test_pipeline.py` exercising `run_pipeline()` directly with mock stages. ADR-0001 names testability as a goal but no current test calls the orchestrator without going through `aggregate()`.

### DIFF
--- a/src/tsam/pipeline/__init__.py
+++ b/src/tsam/pipeline/__init__.py
@@ -9,6 +9,7 @@ from typing import TYPE_CHECKING
 import numpy as np
 import pandas as pd
 
+from tsam.config import ClusteringResult, MinMaxMean
 from tsam.options import options
 from tsam.pipeline.accuracy import reconstruct
 from tsam.pipeline.clustering import (
@@ -28,10 +29,7 @@ from tsam.pipeline.types import (
 )
 
 if TYPE_CHECKING:
-    from tsam.config import (
-        Distribution,
-        MinMaxMean,
-    )
+    from tsam.config import Distribution
 
 
 # ──────────────────────────────────────────────────────────────────────
@@ -148,8 +146,6 @@ def _build_representation_dict(
     cluster_representation: str | Distribution | MinMaxMean | None,
 ) -> dict[str, str]:
     """Build the representation dict (mean/min/max per column) from config."""
-    from tsam.config import MinMaxMean
-
     representation_dict: dict[str, str] = dict.fromkeys(columns, "mean")
     if isinstance(cluster_representation, MinMaxMean):
         for col in cluster_representation.max_columns:
@@ -181,8 +177,6 @@ def run_pipeline(
     Replaces create_typical_periods() + predict_original_data() +
     accuracy_indicators() from the v3 class-based API.
     """
-    from tsam.config import ClusteringResult as _ClusteringResult
-
     cluster = cfg.cluster
     cluster_representation = cluster.get_representation()
     representation_dict = _build_representation_dict(
@@ -410,7 +404,7 @@ def run_pipeline(
         else None
     )
 
-    clustering_result = _ClusteringResult.from_pipeline(
+    clustering_result = ClusteringResult.from_pipeline(
         cluster_center_indices=cluster_center_indices,
         extreme_periods_info=extreme_periods_info,
         extremes_config=cfg.extremes,
--- a/src/tsam/pipeline/__init__.py
+++ b/src/tsam/pipeline/__init__.py
@@ -89,39 +89,15 @@ def _warn_if_out_of_bounds(
             )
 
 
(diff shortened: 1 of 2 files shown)


FBumann:
@julian-belina Does this match what you are aiming for? Just do a quick scann.

julian-belina:
@FBumann, thanks for your quick response! This goes in the right direction. It would be nice if the nomenclature in the pipeline diagram and the orchestrator steps align closely. But this would just be the cherry on top. It also doesnt matter if the diagram or the function names change.

FBumann:
Maybe we should merge the docs pr first before iterating on the code then. I'll review it purely on correctness tomorrow

julian-belina:
sounds good

FBumann:
Closed for now. Will be reconsidered in a PR colocating documentation in the code docstrings.

FBumann:
Closed for now. Will be reconsidered in a PR colocating documentation in the code docstrings.

PR #330, opened 2026-05-29
fix: warn that aggregate() column order will change in v4

FBumann:
## What

In v3, `aggregate()` returns `cluster_representatives`, `reconstructed`, and `original` with columns **sorted alphabetically** (inherited from the legacy `TimeSeriesAggregation` path). In v4 the new pipeline will preserve the **input DataFrame's column order** instead.

This can break downstream code **silently**: anything that reads results by position (`.values`, `.to_numpy()`, `.iloc[:, 0]`, matrix/solver input) keeps the same column names and shape, so nothing raises — but each column's data lands in a different slot.

## Change

`aggregate()` now emits a `FutureWarning`, fired **only when the input columns are not already alphabetical** — i.e. exactly the inputs whose output actually changes. Alphabetical inputs stay silent. The message tells users how to:
- keep the current order: `tsam.aggregate(data.sort_index(axis=1), ...)`, or `result.cluster_representatives.sort_index(axis=1)`;
- adopt the v4 order now: index results by column name (this also silences the warning).

Also:
- Prominent heads-up admonition in the migration guide.
- `Warns` section added to the `aggregate()` docstring.
- New tests in `test/test_column_order_warning.py`.

## Note

`tuning` calls `aggregate()` internally, so tuning users see this warning once, located at the internal call site rather than their own code. Left as-is to avoid a silent path; can refine if preferred.

The legacy `TimeSeriesAggregation` class is unaffected — it keeps sorting alphabetically in both v3 and v4.

### DIFF
--- a/docs/migration-guide.md
+++ b/docs/migration-guide.md
@@ -6,6 +6,33 @@ and will be removed in a future release.
 
 This guide covers every change you need to make.
 
+## Heads-up: column order changes in v4 { #v4-column-order }
+
+!!! warning "`aggregate()` result column order will change in v4"
+
+    In **v3**, `aggregate()` returns `cluster_representatives`, `reconstructed`,
+    and `original` with columns **sorted alphabetically**; in **v4** they follow
+    the **input DataFrame's order**. This can break code that reads results *by
+    position* (`.values`, `.iloc[:, 0]`) **silently** — names and shape are
+    unchanged, but each column's data lands in a different slot. Indexing *by
+    name* is unaffected. `aggregate()` emits a `FutureWarning` when input columns
+    are not already alphabetical (the only case that changes).
+
+    To keep the v3 order, sort before — `aggregate(data.sort_index(axis=1), ...)`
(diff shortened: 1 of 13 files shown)


julian-belina:
Since the behaviour cannot be changed in the current implementation, would it be possible to add a flag to disable the warning?

FBumann:
Where would you like to put that?
I'd add a note to the docs on how to deactivate it directly with warnings.

julian-belina:
Good question. I'm not sure. Perhaps it could be an option for the aggregate function or an attribute in the main Tsam __init__.py that can be set after the initial import.

FBumann:
I would suggest:

```python
import warnings

warnings.filterwarnings("ignore", category=FutureWarning, module="tsam", message=".*sorted.*")

```

FBumann:
That would be much more maintainable. Users would paste it into their code. And in v4 it just does nothing. We don't need to deprecate a flag we just added in v4

FBumann:
@julian-belina Updated the notbeooks and the warning message now returns how to deactivate it

julian-belina (approved):

PR #338, opened 2026-05-31
refactor!: reorganize package structure and drop v3 compat shims

FBumann:
## Summary

Reorganizes `src/tsam/` to clarify two boundaries — **public vs. internal**, and **`algorithms/` (config-agnostic numerical kernels) vs. `pipeline/` (config-aware orchestration)** — and removes the remaining v3 backward-compat shims now that v4 is a breaking release.

## What changed

- **`utils/` → `algorithms/`**, holding the numerical kernels; `ClusteringResult` moved to `result.py` (it's a result type, and the move breaks a `config ↔ result` cycle); shared serialization/inference helpers extracted to a new `commons.py`.
- **Parallel kernel/stage naming**: `algorithms/clustering.py` ↔ `pipeline/clustering.py`, `algorithms/segmentation.py` ↔ `pipeline/segmentation.py`.
- **Removed v3 shims** and dead legacy-name handling.

## Breaking changes

- `AggregationResult.cluster_weights` → `cluster_counts`; `plot.cluster_weights()` → `plot.cluster_counts()` (alias kept, warns).
- `ClusterConfig(normalize_column_means=)` → `scale_by_column_means`; `ClusterConfig(weights=)` removed — pass `weights` to `aggregate()` directly (now a keyword-only argument).
- Legacy verbose representation names and legacy cluster-method names (e.g. `k_means`) are no longer accepted; use the short names (`mean`, `kmeans`, …).
- `tsam.weights.MIN_WEIGHT` → `tsam.options.min_weight`.

See the migration guide for the full v3 → v4 mapping. Per-commit detail is in the commit list below.

_Aggregation **results are unchanged** (golden regression suite passes throughout)_


docs/scripts/check_api_reference.py
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/env python
```
julian-belina, line 1: If possible, I think this script should be moved to the docs folder.

src/tsam/algorithms/clustering.py
```diff
@@ -0,0 +1,131 @@
...
+    "averaging": "mean",
+    "kmeans": "mean",
+    "kmedoids": "medoid",
+    "kmaxoids": "maxoid",
+    "hierarchical": "medoid",
+    "contiguous": "medoid",
+}
+
+
+def assign_clusters(
```
julian-belina, line 25: Currently there is no consistent docstring style. This function annotates arguments inline but does not explain the parameters in the docstring. This can be okay for private functions, but it's not marked as private (e.g. by using a leading underscore).

On the other hand, e.g. _k_medoids_exact only describes types in docstrings and does not use inline annotations. It also documents each parameter in the docstring even though it's private.

I would be open to a convention proposal, but can also suggest one if wanted. I also think this is a bigger change and should probably be done in a separate PR.

src/tsam/plot.py
```diff
@@ -246,7 +246,7 @@ def cluster_members(
         unstacked = unstack_to_periods(result.original, n_ts * timestep_hours)
         assignments = result.cluster_assignments
         representatives = result.cluster_representatives
-        weights = result.cluster_weights
+        weights = result.cluster_counts
```
julian-belina: This should be renamed, to avoid confusion with the user supplied weights

src/tsam/algorithms/clustering.py
FBumann: Agreed fully.
I have a pretty strong opinion on that, and I would keep it as a follow up PR, sth like 4.0.1

src/tsam/plot.py
julian-belina: Same problem in the notbook "optimiztation_input.ipynb"

julian-belina:
Otherwise this looks good, thank you for your contribution.

PR #357, opened 2026-06-17
fix: plot cluster_representatives for segmented results (#340)

### LINKED ISSUE #340: result.plot.cluster_representatives() raises ValueError on segmented (v4) results
Calling `result.plot.cluster_representatives()` on an `AggregationResult` produced **with segmentation** raises a `ValueError` instead of plotting.

## Reproduce

```python
import pandas as pd
import tsam
from tsam import SegmentConfig

raw = pd.read_csv("docs/notebooks/testdata.csv", index_col=0, parse_dates=True)
data = raw.loc["2010-01-01":"2010-02-11"]

result = tsam.aggregate(
    data, n_clusters=6, period_duration="1D",
    segments=SegmentConfig(n_segments=6),
)
result.plot.cluster_representatives(columns=["Load"])
```

```
ValueError: Length mismatch: Expected axis has 4 elements, new values have 3 elements
```

## Cause

`cluster_representatives` assumes the representatives have a **2-level** index (period, timestep):

```python
# src/tsam/plot.py
df = typ[columns].reset_index()
df.columns = pd.Index(["Period", "Timestep", *columns])   # <-- assumes 2 index levels
```

With segmentation the index has **3 levels**:

```
reps index names: [None, "Segment Step", "Segment Duration"]
```

so `reset_index()` yields one extra column and the hardcoded `df.columns` assignment fails.

## Notes
- Non-segmented results plot fine; the bug is specific to segmented aggregations.
- `result.plot.cluster_members()` handles segmentation correctly (it expands segments via the durations), so it can serve as a reference for the fix.
- Surfaced while rewriting the docs; the Quickstart works around it by using `cluster_members` on the segmented example.
- Version: `3.3.1.dev56+gc5ab18654` (v4 functional API).

## Suggested fix
In `cluster_representatives`, build the column names from the actual index (e.g. use the real `reset_index()` column count / collapse the segment levels to a single timestep axis using `Segment Duration`) rather than hardcoding `["Period", "Timestep", *columns]`.

gaoflow:
Fixes #340.

### Problem

`result.plot.cluster_representatives()` raises `ValueError: Length mismatch: Expected axis has 4 elements, new values have 3 elements` on segmented (v4) results. `ResultPlotAccessor.cluster_representatives` assumes the representatives DataFrame has a 2-level `(period, timestep)` index and hardcodes the `reset_index()` column names to `["Period", "Timestep", *columns]`. With segmentation the index has 3 levels `(period, segment step, segment duration)`, so `reset_index()` yields one extra column and the assignment fails.

### Fix

Derive the index-level names from `typ.index.nlevels`: keep `Period`/`Timestep` for the first two levels and drop any extra segment-metadata levels, so the column assignment matches the actual width. Also pass `value_vars=columns` to `melt` so the dropped metadata columns can't leak into the plotted values. The segment x-axis becomes the segment step, consistent with `cluster_members`.

### Tests

Added a `TestClusterRepresentatives` segmentation case using the existing `result_segmented` fixture: the plot no longer crashes and the representative trace y-values match `cluster_representatives.loc[...]`. Without the fix both new tests fail with the length-mismatch; with it `test_plot.py` passes (35) and the full suite is green.

Disclosure: I used AI assistance (Claude) to help locate and draft this fix, under my direction and review.

### DIFF
--- a/src/tsam/plot.py
+++ b/src/tsam/plot.py
@@ -166,15 +166,26 @@ class ResultPlotAccessor:
             columns, available_columns, "cluster_representatives"
         )
 
-        # Reset index to get period/timestep as columns
+        # Reset index to get period/timestep as columns. The representatives
+        # index has 2 levels normally (period, timestep) but 3 with
+        # segmentation (period, segment step, segment duration). Name the
+        # first two levels and drop any extra segment-metadata levels so the
+        # column assignment matches the actual number of columns.
         df = typ[columns].reset_index()
-        df.columns = pd.Index(["Period", "Timestep", *columns])
+        n_index_levels = typ.index.nlevels
+        index_names = [
+            "Period",
+            "Timestep",
+            *[f"_extra_{i}" for i in range(n_index_levels - 2)],
+        ]
+        df.columns = pd.Index([*index_names, *columns])
 
         # Map period IDs to labels with weights
         df["Period"] = df["Period"].map(lambda p: f"Period {p} (n={weights.get(p, 1)})")
 
         long_df = df.melt(
             id_vars=["Period", "Timestep"],
+            value_vars=columns,
             var_name="Column",
             value_name="Value",
         )
--- a/test/test_plot.py
+++ b/test/test_plot.py
@@ -70,6 +70,22 @@ class TestClusterRepresentatives:
         fig = result.plot.cluster_representatives(columns=[col])
         assert isinstance(fig, go.Figure)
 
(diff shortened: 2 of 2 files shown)


jo-omoyele (approved):
Looks good. This removes the hard-coded assumption that representatives always have a 2-level index after `reset_index()`. Segmented representatives add an extra index level, so deriving the index column names from `typ.index.nlevels` avoids the column-length mismatch.

PR #364, opened 2026-06-22
chore(ci): keep mypy in lockstep across all sources via one conda-triggered PR

julian-belina:
## What
Makes Renovate manage the `mirrors-mypy` pre-commit hook through the **conda-forge** datasource (like `pyproject.toml` and `environment.yml` already are) instead of the GitHub tag, and disables the built-in `pre-commit` manager for that one hook so the rev isn't double-managed.

## Why
Today mypy lives in three places under two managers:

| Source | Manager | Trigger |
|---|---|---|
| `pyproject.toml` / `environment.yml` | `custom.regex` → conda-forge | conda release |
| `.pre-commit-config.yaml` (`mirrors-mypy` rev) | built-in `pre-commit` → GitHub tag | GitHub tag |

Different managers, different triggers → they drift. That's how the hook got stuck at `v1.20.2` while the project moved to `2.1.0` (the bump in #308 was closed, Renovate never re-raised it), which broke CI on every open dependency PR. See #363 for the immediate fix.

## How this fixes it
The new `custom.regex` manager uses `depName: mypy`, so the existing grouping rule (`matchManagers: [pep621, custom.regex]`, `groupName: {{depName}}`) folds the hook into the **single `mypy` PR** with the other two sources.

Resolving the hook via `conda-forge/mypy` means that combined PR only opens **once mypy is released on conda-forge** — the slowest of the three channels — which guarantees the matching `mirrors-mypy vX.Y.Z` tag and PyPI release already exist. No race, no partial bumps: all three move together or not at all.

The literal `v` tag prefix is matched *outside* the `currentValue` capture, so it's preserved on write-back (`mirrors-mypy` tags are `vX.Y.Z`; conda versions are `X.Y.Z`).

Validated locally with `renovate-config-validator` (passes).

## Note for reviewers
Stacked on #363 (so CI here runs against the fixed hook and stays green). Please **merge #363 first**; GitHub will then auto-retarget this PR to `develop`. The diff here is `.github/renovate.json` only.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -51,6 +51,20 @@
             "packageNameTemplate": "conda-forge/{{{depName}}}",
             "datasourceTemplate": "conda",
             "versioningTemplate": "pep440"
+        },
+        {
+            "customType": "regex",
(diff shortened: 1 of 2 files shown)


julian-belina:
Superseded by #363. Switching mypy to a `repo: local` hook driven by the project's environment makes pyproject.toml the single source of truth for mypy/numpy/pandas-stubs — there's no longer a hook `rev` or `additional_dependencies` to keep in lockstep, so this Renovate plumbing isn't needed.

PR #373, opened 2026-07-01
fix: Conurrency and integral fix

l-kotzur:
# Concurrency-preserving distribution representation + integral / min-max fixes

Branch: `concurrency_in_representation` → `master`

## Summary

The distribution (duration-curve) representation reproduces each attribute's set
of values ("distribution") but has to decide **how to lay those values out in
time**. The old behaviour ordered each attribute independently, which fits every
marginal distribution best but destroys the *joint* structure across attributes
("concurrency") — e.g. it can place peak solar and peak demand at the same
timestep even when they never co-occur.

This branch makes that ordering configurable, adds strategies that preserve
concurrency at **no cost to the marginal distribution**, adds a metric to measure
it, and fixes the duration representation so it preserves the **integral** and
**min/max** of the original series.

## Concepts

- **Distribution** — each attribute's marginal set of values (its duration
  curve). Preserving it means the model sees the correct range/frequency of
  values per attribute.
- **Concurrency** — the joint structure across attributes: which values co-occur
  at the same timestep (formally the empirical copula / cross-correlation).

All ordering strategies below preserve the **distribution identically**; they
differ only in how much **concurrency** they preserve.

### Ordering strategies (`concurrency`)

| Strategy | Behaviour | Concurrency |
|----------|-----------|-------------|
| `independent` *(default)* | each attribute ordered by its own mean profile | none across attributes |
| `reference` | one attribute's ordering broadcast to all (needs `reference_attribute`) | with the reference attribute |
| `medoid` | every attribute ordered by the cluster **medoid**'s own ranks (a real period) | across **all** attribute pairs |
| `consensus` | one shared ordering from the 1st principal component of all mean profiles | balanced, single shared axis |
| `assignment` | optimal single shared ordering minimising deviation from the cluster mean profile | balanced, single shared axis |

Only valid for cluster-scope (period-wise) distribution. Global-scope
distribution supports `independent` only.

### `concurrencyIndicators()`

New method on `TimeSeriesAggregation` returning the Frobenius error of the
Pearson and Spearman cross-attribute correlation matrices (original vs
reconstructed) — a direct measure of how well concurrency is preserved.
`accuracyIndicators().RMSE_duration` remains the distribution-side metric.

### Benchmark

`benchmark/concurrency_representations.py` compares all strategies on the example
dataset. On `docs/notebooks/testdata.csv` (GHI, T, Wind, Load), every strategy
has identical distribution error (`RMSE_duration = 0.006806`) and:

| strategy | corr_frobenius | spearman_frobenius |
|----------|---------------:|-------------------:|
| `medoid` | **0.258** | **0.229** |
| `independent` (default) | 0.734 | 0.754 |
| `reference` / `consensus` / `assignment` | 1.224 | 1.368 |

`medoid` reduces the concurrency error ~3× below the default at no distribution
cost. The single-shared-axis strategies (including the pre-existing `reference`)
score worse than the default because forcing one ordering onto every attribute
makes them all co-monotone, inventing spurious positive correlation.

## Fixes: integral & min/max preservation

- The duration representation now preserves the **integral by construction**,
  including the `distribution_minmax` case. Min/max pinning redistributes the
  endpoint delta across the interior points (`_pinMinMaxPreserveSum`, bounded
  water-fill within `[min, max]`) instead of shifting the sum — so the integral
  holds **even with rescaling disabled**, and the peaks survive rescaling.
  The single-value segment case (`T == 1`) prioritises the integral, since one
  value cannot carry both min and max.
- `_rescaleClusterPeriods` uses a bounded water-fill that keeps values inside the
  input envelope while restoring the integral.

## API

Typed (recommended):

```python
from tsam import aggregate, ClusterConfig, Distribution

aggregate(df, n_clusters=8, cluster=ClusterConfig(
    method="hierarchical",
    representation=Distribution(
        concurrency="medoid",      # independent | reference | medoid | consensus | assignment
        preserve_minmax=True,      # optional; keeps per-attribute peaks
        # reference_attribute="Load",  # required only for concurrency="reference"
        # scope="cluster",             # "cluster" (default) or "global"
    ),
))
```

Legacy `TimeSeriesAggregation`:

```python
TimeSeriesAggregation(
    df, ...,
    representationMethod="distributionRepresentation",   # or "distributionAndMinMaxRepresentation"
    representationConcurrencyMethod="medoid",
    # representationReferenceAttribute="Load",            # for the "reference" strategy
)
```

Inputs are validated with clear errors (unknown strategy, non-distribution
method, global scope, `reference` without a reference attribute).

## Backward compatibility

**Default behaviour is unchanged** — distribution representations still default
to `independent` ordering; every new strategy is opt-in. The `reference_attribute`
feature is unchanged and equivalent to `concurrency="reference"`.

## Testing

- `test/test_concurrencyRepresentation.py` and
  `test/test_referenceAttributeRepresentation.py` cover every strategy: identical
  marginals, improved concurrency, min/max + integral preservation (including
  without rescaling), config (de)serialisation round-trips, and validation
  errors.
- **Golden regression is green.** Golden files affected by the integral/min-max
  fix were regenerated (`contiguous*`, `hierarchical_distribution_minmax*`,
  `*maxoid*`, `*samemean*` — real value changes, verified as expected). New
  golden cases were added for the new strategies:
  `hierarchical_distribution_{medoid,consensus,assignment,reference}` and
  `hierarchical_distribution_medoid_minmax`. Regenerate with
  `pytest test/test_golden_regression.py --update-golden`.

## Files

- `src/tsam/utils/durationRepresentation.py` — ordering strategies
  (`_orderingForCluster`), sum-preserving min/max (`_pinMinMaxPreserveSum`).
- `src/tsam/representations.py`, `periodAggregation.py`,
  `timeseriesaggregation.py` — plumbing + `concurrencyIndicators()`.
- `src/tsam/config.py`, `api.py` — `Distribution.concurrency` option.
- `benchmark/concurrency_representations.py`, `benchmark/medoid_explainer.html`.
- `test/test_concurrencyRepresentation.py`,
  `test/test_referenceAttributeRepresentation.py`, `test/_golden_cases.py`
  (+ new golden CSVs).


FBumann:
THis is a breaking change I'd say, as the golden regression files where updated.
Also, `test/test_clustering_e2e.py::TestClusteringE2E::test_cluster_representatives[contiguous_medoid_8clusters]` fails, which means its either non-deterministic, or depends on the system (like we had some old code paths do (sorting/ordering/tie-breaking).

Again, id really like to pull this in, but rather after v4 releases #234
@julian-belina

julian-belina:
@l-kotzur Thank you very much for your contribution, and I apologise for the long delay in reviewing it.  The options are a valuable addition, and I will try to incorporate them into the v4 TSAM guide.

I fixed the '--update-golden-files' option, which was missing some tests, and reran it. I also disabled the future warning when running the Pytest suite and set LF as the standard for the end of line symbols.

Before finally releasing v4, I would like to do one last 3.x release, since Fine will only support the new TSM version by the end of the year. However, I would be interested to know if @l-kotzur is already comfortable using v4. 

I will keep a legacy v3 branch live in case any issues are noticed during or after the v4 release.

julian-belina (approved):

FBumann:
@julian-belina Unfortunately this merge introduced a lot of conflicts in #234 
Is this needed? Do you plan to release this?

julian-belina:
@FBumann Unfortunately, yes, this is needed.  [Fine ](https://github.com/FZJ-IEK3-VSA/FINE) will only support Tsam v4x by the end of the year. 

However, the next release should be a 4.0.0 release

FBumann:
Ok. Ill try to resolve it!

julian-belina:
Thank you very much!

julian-belina:
@FBumann, maybe wait another hour. I have a significant PR incoming from the branch 'improve_reworked_notebooks'.

Issue #336, opened 2026-05-31
Post-v4 release: drop check_like=True in golden regression and regenerate golden CSVs in input column order


FBumann:
## Background

In v4, `aggregate()` returns DataFrame columns in **input order**, whereas the legacy API (and the stored golden CSVs) use **alphabetical order**. See the "Column order (new API only)" note in the v3→v4 migration guide.

To keep `test/test_golden_regression.py` passing during the transition, the golden comparison currently uses `check_like=True`, which ignores column (and index) ordering:

```python
pd.testing.assert_frame_equal(
    result.reconstructed,
    golden,
    check_names=False,
    check_freq=False,
    check_like=True,  # v4 preserves input column order; golden CSVs are alphabetical
    atol=1e-7,
)
```

This was introduced in #335 (merge of `develop` into `develop-v4`).

## Why this is a temporary measure

`check_like=True` weakens the regression gate: it no longer asserts that columns come back in the expected order. Once v4's input-order behavior is the released contract, the goldens should encode that order so the test verifies it strictly again.

## Task (do this **after v4 is released**)

- [ ] Regenerate all golden CSVs under `test/data/golden/` so they are stored in **input column order** (re-run the golden generation against the v4 pipeline).
- [ ] Remove the `check_like=True` argument from the `assert_frame_equal` call in `test/test_golden_regression.py`.
- [ ] Confirm the suite still passes with strict column-order checking.

## Notes

- The 111 golden CSVs are currently byte-identical between `develop` and `develop-v4`; regeneration will reorder columns only (values unchanged).

PR #376, opened 2026-07-01
fix!: preserve integral and min/max envelope in duration representation (Part 1)

FBumann:
I pulled the pure "bugfix" parts out of #373 and made them fit v4 directly instead of pulling it into v3. 
I made sure the created golden files match#373 exactly, so implementations are equal.

I also made some design decisions:
- no magic numbers: pulled the repeats and tolerance into tsam.options

Further, this changes the results of tsam, in some cases meaningfully. So its a breaking change.

---
AI generate content below

## Summary

Ports the **integral / min-max preservation fix** into the v4 pipeline. This is the bug-fix half of the work in @l-kotzur's PRs #282 and #373, re-implemented against the v4 architecture (the originals target `develop`/v3, where the touched files were renamed or deleted in v4, so they cannot be merged directly).

**This is Part 1 of 2.** Part 2 (the concurrency-preserving distribution ordering + `concurrencyIndicators`) will be stacked on top of this branch.

## The bug

In rare cases the aggregated series violated the original lower/upper bounds — the `At least one maximal value of the aggregated time series exceeds the maximal value the input time series` UserWarning. The min/max representation and the cluster-period rescaling both used an iterated *multiply-and-clip*, which flattens values against a hard cap and distorts the integral.

## The fix

Replace the multiply-and-clip with a single **bounded water-fill**: a shape-preserving warm start, then redistribution of the residual proportional to each element's remaining headroom (or depth when shrinking), keeping every value inside `[min, max]` while restoring the integral.

Extracted as **one shared primitive** `commons.bounded_water_fill`, with three callers:

| Caller | Scope | Status |
|---|---|---|
| `duration_representation._represent_min_max` | global-scope distribution | rewritten |
| `duration_representation._pin_min_max_preserve_sum` | period-wise / cluster scope | **new** |
| `pipeline.rescale.rescale_representatives` | cluster-period rescaling | rewritten |

The new `_pin_min_max_preserve_sum` replaces the naive endpoint overwrite (`repr_values[:, 0] = flat[:, 0]`) that silently shifted the integral. For single-value segments (`T == 1`) it prioritizes the integral, since one value cannot carry both the min and the max.

The pass cap and tolerance move into `tsam.options` (`minmax_max_passes=100`, `minmax_tolerance=1e-12`) next to the existing `rescale_*` knobs.

## Validation

- **All 15 regenerated golden files match PR #373 bit-for-bit** (Δ=0), an independent cross-check that this port reproduces the reference implementation exactly.
- Golden diffs are value-only; input column order is preserved (no premature #336 reorder).
- Relaxed/added `test_durationRepresentation` cases assert the envelope is now respected with and without rescaling.
- Full suite green (415 passed).

## ⚠️ Behavior change → needs release notes

Results shift for peaky columns near the envelope. Per the golden policy (option 1), goldens were regenerated. The changes are concentrated: e.g. `contiguous_medoid` moves only in `Wind` (the one peaky column), max Δ 3.6e-4; larger shifts appear in the `kmaxoids`/`samemean`/`distribution_minmax` cases where clipping was most active — all matching #373.

Closes #282 (superseded here for the v4 line)

---

## Impact analysis (transparency)

I measured old (`develop-v4`) vs this branch on the integral-preservation metric (reconstructed sum vs input sum, worst column). **This branch reproduces PR #373 bit-for-bit on every scenario below**, so the numbers are the reference fix's behavior, not a v4-port artifact.

| representation | rescale | old | this PR |
|---|---|---:|---:|
| cluster-scope min/max | **ON** (default) | 0.00% | 0.00% |
| cluster-scope min/max | OFF | **5.04%** | **0.00%** ✓ |
| min/max + 8 segments | **ON** (default) | **22.5%** | **0.00%** ✓ |
| min/max + 8 segments | OFF | 11.3%† | 14.27%† |

**In normal usage (rescaling on, the default) the fix eliminates large min/max integral errors** — up to 22.5% → 0% for the segmented case — and is neutral where the old code was already correct. With rescaling off it fixes the cluster-scope case (5.04% → 0%).

### The one cell that looks *worse* (seg + rescale OFF: 11.3% → 14.27%) — and why it isn't a regression

That 14.27% is a **pre-existing clustering integral loss**, unrelated to min/max: aggregating 365 days into 8 typical days drops Wind's annual integral by 14.27% even for a **plain mean representation with no min/max** (rescaling is exactly what corrects it — hence 0% when it's on).

The old code *accidentally masked* part of that loss with a latent bug: for a single-value segment (`T == 1`), the inline pinning ran `repr_values[:, 0] = min; repr_values[:, -1] = max` on the **same cell**, leaving it at the **max**. That over-inflated Wind and pushed the under-estimated integral back up → the "better" 11.3%.

The fix does the correct thing — a single value can't carry both min and max, so `_pin_min_max_preserve_sum` leaves `T == 1` segments at the **mean** (integral-neutral, #373's documented rule). It no longer masks the clustering loss, so the honest 14.27% shows through (equal to the base clustering loss to the digit — confirming the min/max step now adds *zero* distortion). **11.3% was the right-looking number for the wrong reason; 14.27% is correct.**

### On the golden suite itself

The 15 regenerated goldens all have rescaling on, where integral and envelope are preserved by **both** old and new — so those value diffs (some large in absolute terms, e.g. Δ180 on a peak cell) are **equivalent-quality distribution redistributions**, not corrections of a broken integral. The correctness gains above show up specifically where rescaling is disabled or where the old min/max pinning distorted the shape.


l-kotzur (approved):

jo-omoyele (approved):
The core fix, replacing the blind endpoint and overwriting with bounded_water_fill, is cool.

julian-belina (approved):

julian-belina:
Thank you for your work, @FBumann, and sorry for the long delay.

PR #377, opened 2026-07-01
feat: concurrency-preserving distribution ordering (Part 2)

FBumann:
## Summary

Ports the **concurrency-preserving distribution representation** from @l-kotzur's PR #373 into the v4 pipeline. This is **Part 2 of 2**, stacked on **Part 1 (#376)** — review/merge #376 first; this PR targets that branch, so its diff shows only the Part 2 additions.

The distribution representation reproduces each attribute's marginal distribution (duration curve) but must decide *how to lay those values out in time*. This makes that choice configurable so cross-attribute **concurrency** (which values co-occur at the same timestep — the empirical copula) can be preserved **at no cost to the marginal distribution**.

## New API

```python
from tsam import aggregate, ClusterConfig, Distribution

aggregate(df, n_clusters=8, cluster=ClusterConfig(
    representation=Distribution(
        concurrency="medoid",   # independent | reference | medoid | consensus | assignment
        # reference_attribute="Load",   # required for concurrency="reference"
        # preserve_minmax=True,
    ),
))

result.concurrency   # Series: corr_frobenius, spearman_frobenius (lower = better)
```

| Strategy | Ordering | Concurrency |
|---|---|---|
| `independent` *(default, unchanged behaviour)* | each attribute by its own mean profile | none across attributes |
| `reference` | one attribute's ordering broadcast to all | with the reference attribute |
| `medoid` | the cluster medoid's per-attribute ranks (a real period) | across all attribute pairs |
| `consensus` | 1st principal component of all mean profiles | single shared axis |
| `assignment` | optimal single ordering vs the cluster mean profile | single shared axis |

On the example data, `medoid` cuts the concurrency error ~3× below the default (`corr_frobenius` 0.258 vs 0.734) at **identical** `rmse_duration` — reproducing #373's benchmark table exactly.

## Architecture (v4-native, not a transliteration of the v3 monolith)

- **`algorithms/concurrency.py`** — new module: a pure strategy registry + `compute_ordering()`. Keeps the ordering concern out of `duration_representation` (which owns the marginal values).
- **name→index** for `reference_attribute` is resolved at the **orchestrator** boundary (mirroring `_build_representation_dict`) and threaded down as one scalar, so the `algorithms/` layer stays index-based and pure.
- `Distribution.concurrency` / `reference_attribute` + validation in `config.py`.
- **`result.concurrency`** metric (Pearson/Spearman correlation Frobenius) computed in `pipeline/accuracy.py`, surfaced as a lazy property next to `result.accuracy`.

## Determinism fixes over #373

Both were latent cross-platform hazards for the goldens:
- **`consensus`** canonicalises the SVD singular-vector sign (LAPACK sign is arbitrary; a flip would reverse the ordering).
- **`medoid`** rounds the summed distances before `argmin` so exact ties resolve identically across BLAS/platforms.

## Guards (concurrency does **not** apply everywhere)

- Requires `scope="cluster"` (period-wise) — rejected for `scope="global"`.
- **Rejected on segment representations** with a clear message (a segment's axis is the segment structure, not per-timestep) — set it on the cluster representation instead.
- Works across all cluster methods, the `use_duration_curves` path, and predefined assignments.

## Validation

- **All 5 new golden cases match #373 bit-for-bit** (Δ=0, column-aligned) — independent proof the port reproduces the reference implementation.
- Marginals (`rmse_duration`) identical across all strategies; the metric reproduces #373's benchmark.
- New `test/test_concurrencyRepresentation.py` (22 tests): marginal invariance, per-strategy ordering semantics, determinism, envelope+integral under min/max, config round-trips, and all validation guards.
- `scipy` added as an explicit dependency (used directly by `assignment`).
- Full suite green; ruff / ruff-format / mypy / api-reference all pass.

## Not included (deliberately, to keep the PR focused)

- No committed benchmark script — the golden cases already pin each strategy's output bit-for-bit, so the numbers can't drift. A single-use replica of #373's table is in the collapsible section below instead.

## Benchmark replica (PR-only, not committed)

The behaviour is already locked by committed tests + golden cases (the golden CSVs pin each strategy's reconstructed output bit-for-bit, so the concurrency/distribution numbers cannot drift). The script below is a **single-use** reproduction of PR #373's published tradeoff table on the v4 API — every value matches, confirming the port. It is intentionally not part of the committed code.

<details>
<summary>Replica script + output (reproduces PR #373's table on v4)</summary>

```python
"""Replica of PR #373's concurrency-representation benchmark, on the v4 API.

Reproduces the published tradeoff table so the v4 port can be spot-checked
against the original. Standalone — not part of the test suite. Run with:

    python concurrency_replica.py

Setup mirrors #373: hierarchical clustering, 8 typical periods, 24 h,
distribution representation, rescaling off, on docs/notebooks/testdata.csv.
"""

import warnings
from pathlib import Path

import pandas as pd

from tsam import ClusterConfig, Distribution, aggregate

DATA = Path("docs/notebooks/testdata.csv")
STRATEGIES = ["independent", "reference", "medoid", "consensus", "assignment"]


def run(raw, method):
    kwargs = {"reference_attribute": "GHI"} if method == "reference" else {}
    result = aggregate(
        raw,
        n_clusters=8,
        period_duration=24,
        cluster=ClusterConfig(
            method="hierarchical",
            representation=Distribution(concurrency=method, **kwargs),
        ),
        preserve_column_means=False,
    )
    conc = result.concurrency
    return (
        result.accuracy.weighted_rmse_duration,
        conc["corr_frobenius"],
        conc["spearman_frobenius"],
    )


def main():
    raw = pd.read_csv(DATA, index_col=0)
    header = f"{'strategy':13s} {'RMSE_duration':>13s} {'corr_frob':>10s} {'spearman_frob':>13s}"
    print(header)
    print("-" * len(header))
    with warnings.catch_warnings():
        warnings.simplefilter("ignore")
        for method in STRATEGIES:
            dist, corr_f, spear_f = run(raw, method)
            label = "reference(GHI)" if method == "reference" else method
            print(f"{label:13s} {dist:13.6f} {corr_f:10.4f} {spear_f:13.4f}")


if __name__ == "__main__":
    main()
```

Output:

```
strategy      RMSE_duration  corr_frob spearman_frob
----------------------------------------------------
independent        0.006806     0.7337        0.7536
reference(GHI)      0.006806     1.2236        1.3677
medoid             0.006806     0.2582        0.2290
consensus          0.006806     1.2236        1.3677
assignment         0.006806     1.2236        1.3677
```

Matches PR #373 exactly — `medoid` 0.258 / 0.229, `independent` 0.734 / 0.754, single-shared-axis (`reference`/`consensus`/`assignment`) 1.224 / 1.368, and `RMSE_duration` 0.006806 for every strategy.
</details>

## Ordering-cost scaling (on-demand, not committed)

Built and run on demand with [`pytest-benchmem`](https://pypi.org/project/pytest-benchmem/) (a drop-in for `pytest-benchmark`) to characterise the cost profile of the ordering strategies — **min time per cluster (µs)**, across strategy × period length T:

| strategy | T=24 | T=48 | T=96 | T=192 | T=384 | T=768 |
|---|---:|---:|---:|---:|---:|---:|
| `independent` | 0.8 | 1.5 | 3.7 | 8.1 | 19.0 | 47.0 |
| `reference` | 2.0 | 2.1 | 2.7 | 3.8 | 6.2 | 12.5 |
| `consensus` | 26.5 | 29.6 | 32.9 | 38.4 | 52.0 | 74.2 |
| `medoid` | 92.6 | 127.0 | 297.6 | 556.8 | 1,174.2 | 2,556.1 |
| `assignment` | 14.9 | 70.7 | 681.2 | 5,420.7 | 39,757.1 | 233,240.4 |

`assignment`'s Hungarian solve is **O(T³)** (~6–9× per doubling of T) → impractical for long periods, whereas `consensus`/`reference` stay ~flat. `medoid` grows with cluster size (the pairwise-distance matrix), not shown here. This is the basis for the docstring steer — `medoid` for best concurrency, `assignment` for small-T single-axis, `consensus`/`reference` for long periods.

<details>
<summary>benchmark script (pytest-benchmem, drop-in for pytest-benchmark)</summary>

```python
"""On-demand scaling benchmark for the concurrency ordering strategies.

Times the per-cluster ordering step across two axes — strategy and period
length T — to expose the cost profile (notably assignment's O(T^3) Hungarian
solve vs the near-flat cost of the other strategies).

Run:  pytest bench_concurrency.py --benchmark-only --benchmark-json=bench.json
"""

import numpy as np
import pytest

from tsam.algorithms.concurrency import compute_ordering

STRATEGIES = ["independent", "reference", "medoid", "consensus", "assignment"]
PERIOD_LENGTHS = [24, 48, 96, 192, 384, 768]

N_ATTRS = 4
N_CANDS = 30


@pytest.mark.parametrize("method", STRATEGIES)
@pytest.mark.parametrize("t", PERIOD_LENGTHS, ids=lambda t: f"T{t:04d}")
def test_ordering_cost(benchmark, method, t):
    rng = np.random.RandomState(0)
    cluster_data = rng.rand(N_ATTRS, N_CANDS, t)
    means = np.round(cluster_data.mean(axis=1), 10)
    repr_values = np.sort(cluster_data.reshape(N_ATTRS, -1), axis=1)[:, :t]
    reference_idx = 0 if method == "reference" else None
    benchmark(compute_ordering, cluster_data, means, repr_values, method, reference_idx)
```

Run: `pytest bench_concurrency.py --benchmark-only` (add `--benchmark-memory` for memray peak memory, or `benchmem plot <json> --view scaling --x t --color method` for the log-log scaling plot).
</details>


l-kotzur:
Thanks a lot for digging into it and the clean up.

l-kotzur (approved):

FBumann:
@l-kotzur One more change:
changed the error names or `correlation_error` and `rank_correlation_error`.

FBumann:
@julian-belina I think this didn't land in the pipeline...
It was merged into another branch that was merged BEFORE this was merged...

I'll check it later

PR #379, opened 2026-07-01
docs: normalize docstrings to Google style (#339)

### LINKED ISSUE #339: Normalize docstrings
We currently don't have a clean convention on how docstrings should look like.

I propose this convention:

- Google style
- Short Explanation of the class/merhod/func/property
- function parameters 
- no types in the docstrings. Rely on type annotations. Also for return types

```python
def function_with_pep484_type_annotations(param1: int, param2: str) -> bool:
    """Example function with PEP 484 type annotations.

    Args:
        param1: The first parameter.
        param2: The second parameter.

    Returns:
        The return value. True for success, False otherwise.

    """

```

Sources:

https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

https://google.github.io/styleguide/pyguide.html#s3.8.1-comments-in-doc-strings

FBumann:
Closes #339.

Converts all docstrings in `src/tsam` to **Google style** and removes redundant information that already lives in the signature.

This is a **docstring-only** change — the AST of every touched file is identical to `develop-v4` aside from docstrings, so behavior is unchanged. Verified: ruff check + format, mypy, and the doctest set all match baseline (0 new failures).

### Convention
- **Google style** (`Args`/`Returns`/`Raises`/`Attributes`/`Note`/`Examples`) — no NumPy `Parameters\n----------` or RST `:param:`/`:type:` blocks.
- **No types in docstrings** — types live in the signature annotations (docs render them via `show_signature_annotations: true`).
- **No defaults in docstrings when visible in the signature** — drop `Defaults to X.` / `(default: X)` when the param/field already shows `= X`. Kept when conditional/computed (e.g. `ClusterConfig.representation` "Default depends on method: ...") or not in a signature (`Options` `@property` defaults; a title resolved as `title or "..."`).
- **`Args:` vs `Attributes:`**
  - `Args:` for classes the user **instantiates** (config/estimator classes: `ClusterConfig`, `SegmentConfig`, `ExtremeConfig`, `Distribution`, `MinMaxMean`, `KMedoids`, `KMaxoids`) — documented names are the constructor params, so IDEs show call-site hints.
  - `Attributes:` for classes that **carry results/stats** (`AggregationResult`, `AccuracyMetrics`, `ClusteringResult`, `TuningResult`, internal pipeline containers) — these document readable attributes and computed properties that are not constructor arguments.

The convention is also written up in `CONTRIBUTING.md` (Code Style → Docstrings) with links to the Google style guide and the napoleon example.


julian-belina:
Using only type annotations and not repeating them in the docstrings is a good approach. However, some functions were missing type annotations completely, while others only stored them only in the docstrings. Therefore, I would be grateful if you could add type annotations for all functions, as I find these very helpful when reviewing.

FBumann:
Can we merge this anyway? And do a follow up up? Or do you need it all at once?

julian-belina:
Ideally, we would have it all at once so that we could see if the old docstrings were incorrect.

FBumann:
@julian-belina Ready

FBumann:
@julian-belina I fixed mypy issues and a small merge bug I introduced during our call.
Nothing that should affect your review

julian-belina (approved):

julian-belina:
@FBumann Should we squash or merge here?

FBumann:
EDIT: MERGE. Else the other branch has conflicts

Issue #378, opened 2026-07-01
preserve_minmax is silently ignored for scope='cluster' segment representations


FBumann:
## Summary

When a `Distribution` is used as a **segment** representation, `preserve_minmax=True` only has an effect with `scope="global"`. With `scope="cluster"` (the default) it is **silently ignored** — no error, no warning, no change in output.

## Why

Segmentation represents each segment by a **single value per attribute** (the representation runs with `n_timesteps_per_period=1`; see `algorithms/segmentation.py`). In the period-wise (`scope="cluster"`) path, `_pin_min_max_preserve_sum` hits its `T < 3` guard and returns the integral-preserving mean unchanged — a single value cannot carry both the min and the max, so the integral is (correctly) prioritized. The `scope="global"` path goes through `_represent_min_max`, which does pin endpoints, so there `preserve_minmax` works.

This behavior is **correct** (integral preservation at `T == 1` is the right call), but it is **silent**: a user who sets `preserve_minmax=True` on a cluster-scope segment representation gets no effect and no signal.

## Evidence

```python
# scope='cluster': preserve_minmax has no effect (result identical)
seg_cluster        = Distribution(scope="cluster")
seg_cluster_minmax = Distribution(scope="cluster", preserve_minmax=True)   # -> same output, Δ=0

# scope='global': preserve_minmax does take effect (max 672 -> 731)
seg_global         = Distribution(scope="global")
seg_global_minmax  = Distribution(scope="global", preserve_minmax=True)    # -> changes output
```

## Proposed fix

Emit a `UserWarning` (or raise) when `preserve_minmax=True` is set on a `scope="cluster"` **segment** representation, so that no `Distribution` option is ever silently ignored — symmetric with how `concurrency` / `reference_attribute` are already guarded for segment representations. The interaction is now documented in `SegmentConfig.representation` (PR #377), so this is purely about the runtime signal.

## Notes

- Pre-existing behavior (not a v3 regression), surfaced while reviewing the segment × `Distribution` interaction for the concurrency feature (#377). Non-blocking for v4.
- Touches the Part 1 min/max path (#376), so it is kept out of the #377 (concurrency) PR deliberately.

FBumann:
Resolved on develop-v4: the runtime `UserWarning` for `preserve_minmax` on a local/cluster-scope segment representation is implemented in `config.py` (SegmentConfig.__post_init__), and the interaction is documented via #383. Closing as done for v4.

PR #380, opened 2026-07-01
refactor: rename Distribution scope "cluster" → "local" (deprecate alias)

FBumann:
## Summary

Renames `Distribution(scope="cluster")` → `Distribution(scope="local")`. `"cluster"` is a **misleading name** for **segment** representations, where the group whose distribution is preserved is a *segment*, not a cluster.

The knob is stage-neutral, so the name should be too:
- **`"local"`** — preserve each *group's own* distribution. The group is a **cluster of periods** for a cluster representation, or a **segment of timesteps** for a segment representation.
- **`"global"`** — preserve only the **enclosing whole's** distribution (the full time series for a cluster rep; a single period for a segment rep). Unchanged.

## Backwards compatibility

Fully backwards compatible:
- `"cluster"` is still accepted as a **deprecated alias**, normalized to `"local"` in `Distribution.__post_init__` with a `DeprecationWarning`.
- `to_dict` / `from_dict` and the internal `period_wise` mapping use the new value.
- A behavioural-equivalence test asserts `scope="cluster"` produces **bit-for-bit identical** aggregation output to `scope="local"`.

Serialization stays clean: `to_dict` omits `scope` when it equals the default, so existing JSON (which only ever stored `"global"` explicitly) is unaffected.

## Why now

`develop-v4` hasn't released, so renaming costs no downstream deprecation cycle — this is the cheapest moment to fix the name, while keeping the alias for any in-flight code.

## Not in scope

This is a pure, no-behaviour-change rename. The silent `preserve_minmax` no-op for `scope="local"`/`"cluster"` **segment** representations (#378) is a separate, orthogonal concern and will be a follow-up on top of the concurrency PR (#377), where the segment guard lives.

## Test plan
- [x] New `test/test_distribution_scope.py`: default is `"local"`; `"cluster"` warns + normalizes; `to_dict`/`from_dict` round-trips; behavioural equivalence of the alias.
- [x] Existing representation, serialization, and golden-regression suites pass (200 passed, 111 skipped).

### DIFF
--- a/docs/migration-guide.md
+++ b/docs/migration-guide.md
@@ -165,7 +165,7 @@ The table below maps every old parameter to its v3 equivalent.
 | `addPeakMin` | `ExtremeConfig(min_value=...)` | |
 | `addMeanMax` | `ExtremeConfig(max_period=...)` | |
 | `addMeanMin` | `ExtremeConfig(min_period=...)` | |
-| `distributionPeriodWise` | `Distribution(scope="cluster"\|"global")` | See [representation objects](#typed-representation-objects). |
+| `distributionPeriodWise` | `Distribution(scope="local"\|"global")` | See [representation objects](#typed-representation-objects). |
 | `representationDict` | `MinMaxMean(max_columns=[...], min_columns=[...])` | See [representation objects](#typed-representation-objects). |
 
 ### Cluster method values { #cluster-method-values }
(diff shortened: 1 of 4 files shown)


FBumann:
Superseded — re-filing directly from an upstream branch instead of the fork.

PR #387, opened 2026-07-03
Benchmarks: realistic scaling + representation suite and Performance docs page

FBumann:
Adds a realistic-usage benchmark suite for `tsam.aggregate` plus a focused **Performance & Scaling** docs page built from it.

> Stacked on #377 (`v4/concurrency-representation`) — base will be retargeted once that merges. Draft: the docs numbers are from a local Apple-M3 run; open to feedback on levels/wording.

## What's here
- **`benchmarks/bench_scaling.py`** — one parametrized `test_scale` over a fixed **1-year horizon**, with the levers a user actually dials as native pytest dimensions: `resolution` (1h/15min/5min), `period` (daily/weekly), `columns` (4/256/1000), `clusters`, `method`. It *derives* `timesteps`/`W`/`periods`, so the period count (≈365 daily / ≈52 weekly) stays realistic instead of exploding. Also `test_representation` — the representation methods, all clustered with hierarchical.
- **Fast & bounded (~9 min).** One measured run per case; a 3 s per-case budget (measured-monotonic for cheap methods; a 15 s subprocess hard-kill for the hang-prone `kmedoids`/`kmaxoids`) plus a memory budget drop the pathological corners as gaps.
- **`benchmarks/_data.py`** deterministic tiled-`testdata.csv` generator + levels; **`_timeout.py`** subprocess timeout; **`make_docs_figures.py`** renders the two figures via `pytest_benchmem.load_long_df`.
- **`docs/background/performance.md`** — a short, figure-led story: *fast even for large data → why (periods stay bounded) → the one choice that matters (method) → representation is cheap.*
- Wiring: `pytest-benchmem[plot]` dev dep, collect `bench_*.py`; removes the old `benchmarks/bench.py`.

## The story (two figures)
1. **Method scaling** — x = resolution, colour = method, faceted daily vs weekly: the scalable methods stay sub-second even at 5-min resolution; methods differ by orders of magnitude; and *which method is affordable depends on the setting* (kmedoids is infeasible on 365 daily periods but fine on 52 weekly ones).
2. **Representation cost** — all representations scale together within ~2–3×, i.e. representation choice is cheap.

**Rule of thumb surfaced in the docs:** default to `hierarchical`.

## Notes
- Uses `pytest-benchmem` (memray-backed; macOS/Linux) — a new, alpha dependency.
- Benchmark result JSONs live under the gitignored `.benchmarks/`; the two rendered figure HTMLs are committed so the docs render.


FBumann:
No prio here!

FBumann:
Superseeded by #443

PR #403, opened 2026-07-22
chore: write goldens with alphabetical columns to de-noise the #234 diff

FBumann:
## Why

The umbrella #234 (`develop-v4` → `develop`) diff showed **~87,000 lines** of
golden churn across 5 concurrency goldens. It looked alarming but was **pure
column reordering, zero value change** — v4 preserves input column order
(`GHI,T,Wind,Load`) while the v3 line stores columns alphabetically
(`GHI,Load,T,Wind`). Verified cell-by-cell: `max|Δ| = 0.0`.

This PR de-noises that diff so the reviewer can trust #234 at a glance.

## What

- One line in `_save_golden`: `df.round(8).sort_index(axis=1).to_csv(path)`.
- Regenerated the goldens (`pytest test/test_golden_regression.py --update-golden`).

Result — the #234 golden diff collapses from ~87k lines to **144 lines** across
3 files (`hierarchical_mean` ×2, `kmeans_weighted`), every one an **8th-decimal
rounding refresh** (`max|Δ| = 1e-8`, 0 cells above the 1e-7 test tolerance). The
5 concurrency goldens now byte-match develop.

`check_like=True` stays because live output is still input-order.

## Temporary — reverted after release

This trades away column-order strictness for a clean review diff. The **stacked
follow-up PR** reverts this line, regenerates goldens in v4's native input order,
and drops `check_like=True` (issue #336). **That PR is scheduled to merge only
after v4 is released.**

Golden regression suite: **116 passed**.


FBumann:
Recreating on an upstream branch so the follow-up #336 PR can stack on it cleanly.

PR #405, opened 2026-07-22
chore: [HOLD until v4 release] goldens to native order + drop check_like (#336)

### LINKED ISSUE #336: Post-v4 release: drop check_like=True in golden regression and regenerate golden CSVs in input column order

FBumann:
## ⛔ DO NOT MERGE until v4 is released

Stacked on #403. This is the **revert half** of the golden de-noise, plus the
#336 cleanup. Keep it as a draft until v4 ships — merging it early would put the
~87k-line column-order churn back into the umbrella #234 diff.

## What it does

- Reverts the `sort_index(axis=1)` in `_save_golden` → goldens regenerate in v4's
  **native input column order**.
- **Drops `check_like=True`** from `test_api_matches_golden`, so the golden test
  now asserts column order exactly rather than tolerating it.
- Regenerates all goldens accordingly.

## Why the diff is large

The committed goldens were mostly stored **alphabetically** (a v3 holdover), and
`check_like=True` masked that they didn't match v4's real (input-order) output.
Moving to native order touches ~100 goldens — that's the mismatch finally being
made explicit. No values change beyond 1e-8 rounding.

Full suite (exact comparison, no `check_like`): **455 passed, 117 skipped**.

Closes #336.


FBumann:
Superseded by a rebuild on current develop-v4 (post #404/#406). The base branch here was the now-merged #404 branch.

PR #407, opened 2026-07-22
chore: goldens to v4 native column order + drop check_like (#336)

### LINKED ISSUE #336: Post-v4 release: drop check_like=True in golden regression and regenerate golden CSVs in input column order

FBumann:
## What it does (the #336 cleanup)

- Reverts the temporary `sort_index(axis=1)` in `_save_golden` → goldens are
  written in v4's **native (input) column order**.
- **Drops `check_like=True`** so the golden test asserts column order exactly
  instead of tolerating a mismatch.
- Regenerates all goldens accordingly (128 files).

## Why now

The alphabetical sort (#404/#406) was a review aid: it kept the umbrella #234
diff free of column-order churn while v4 was still in review. **v4.0.0 is
released** and the v4 line is merged into `develop`, so the goldens go back to
recording what the pipeline actually emits.

## Safety of the golden churn

All 128 golden diffs are **pure column reorderings** — verified by comparing
each file before and after with columns sorted on both sides. Every value is
identical; not one number moved.

Full suite (exact comparison, no `check_like`): **695 passed**.

Closes #336.


julian-belina:
I need to hold onto this one until the following pull request is merged: https://github.com/FZJ-IEK3-VSA/FINE/pull/827.

FBumann:
I dont understand why, but I don't mind

julian-belina (approved):

PR #414, opened 2026-07-23
feat(extremes): count-stable clustering and transferable replace (#410)

### LINKED ISSUE #410: Extreme periods are neither count-stable nor transferable — blocks building on top of aggregate()/apply()
# Extreme periods are neither count-stable nor transferable — blocks building on top of `aggregate()`/`apply()`

**tsam version:** 3.4.1 · **Python:** 3.11

## Context

I maintain [`tsam_xarray`](https://github.com/FBumann/tsam_xarray), a thin xarray wrapper around tsam v3. Two very common patterns there run into the same underlying gap in how extreme periods are handled:

1. **Fan-out:** aggregate many independent datasets (scenarios/regions) with the *same* config and stack the results into one array. This needs every aggregation to produce the **same number of clusters**.
2. **Transfer:** compute a clustering once and re-apply it to other datasets via `ClusteringResult.apply()`. This needs the transferred representatives to **match** the ones from a direct `aggregate()`.

`append`/`new_cluster` break (1); `replace` breaks (2). There is no extreme method that satisfies both today, so anything built on top has to reject or work around extremes. Both are avoidable in tsam, and fixing them upstream would help every downstream user, not just this wrapper.

Self-contained repro (tsam only) at the bottom.

---

## Problem 1 — the number of appended extreme clusters is data-dependent

With `method="append"` (or `"new_cluster"`), the final `n_clusters` is `requested + k`, where `k` is *not* the number of configured extreme criteria — it's however many survive dedup. Same config, 6 different datasets, `n_clusters=8`:

```
seed=0: n_clusters=10      seed=3: n_clusters=10
seed=1: n_clusters=13      seed=4: n_clusters=12
seed=2: n_clusters=11      seed=5: n_clusters=12
```

**Root cause** — the two dedup guards in `_addExtremePeriods` (`timeseriesaggregation.py:729-743`, and the identical min/meanMax/meanMin siblings):

```python
stepNo = groupedSeries[column].max(axis=1).idxmax()
if (
    stepNo not in extremePeriodNo                                   # (a) already an extreme for another criterion
    and groupedSeries.loc[stepNo, :].values.tolist() not in ccList  # (b) period already IS a cluster center
):
    ...
```

Both conditions depend entirely on the data (whether the peak-solar day is also the min-wind day; whether an extreme period happens to be a medoid), so the extra-cluster count is unpredictable across datasets that share a config.

This is reasonable *default* behavior, but there's no way to opt into a **fixed, config-determined count**. That makes it impossible to line up results from independent aggregations without a post-hoc reconciliation pass in every wrapper.

**What would help (any one of these):**
- An option to **guarantee one appended cluster per configured criterion** (skip the dedup, allowing duplicate representative periods) — count becomes `requested + n_criteria`, fully predictable.
- Or: expose, on the result, **which criteria were deduped into which existing cluster** (e.g. a `{criterion: cluster_idx}` map), so callers can deterministically pad/align themselves.
- Or at minimum: document that the count is data-dependent and provide `result.clustering.n_clusters` guidance for the fan-out case.

The first option is the most useful for downstream code.

---

## Problem 2 — `replace` silently drops its peak injection on `apply()`

`replace`'s entire purpose is to preserve peaks by overwriting one column of a cluster center with the extreme period's value. That injection is **not reproduced** when the clustering is transferred via `apply()`. Applying a clustering back onto the *same* data it was built from:

```
method=append      reps identical: True    max|Δrep|=0.000   rmse 0.2931 -> 0.2931
method=new_cluster reps identical: True    max|Δrep|=0.000   rmse 0.2909 -> 0.2909
method=replace     reps identical: False   max|Δrep|=0.966   rmse 0.3104 -> 0.3094
```

The representatives differ by up to `0.97` (on data in `[0, 1)`) — i.e. the injected peaks are gone. Note the transferred RMSE actually *improves* (0.3104 → 0.3094), because the fallback plain-medoid fits the bulk better. So the one guarantee `replace` exists to provide is silently lost, and the headline accuracy metric hides it.

**Root cause** — `apply()` replays stored `cluster_centers` (period *indices*) and never re-detects extremes (`config.py:1060-1063`). `append`/`new_cluster` store real period indices, so they round-trip exactly. `replace` builds a hybrid center (a medoid with one column mutated, `timeseriesaggregation.py:864-866`) that has **no addressable period index**, and `_build_clustering_result` deliberately excludes it from `cluster_centers` (`api.py:472-477`). So on transfer the affected cluster falls back to the plain medoid.

There *is* a `UserWarning` (`config.py:992-1006`) and the `extremes_config` is already stored on the `ClusteringResult` — so `apply()` has everything it needs to redo the injection.

**What would help:**
- Make `apply()` **re-detect and re-inject** `replace` extremes on the new data using the stored `extremes_config` (it already re-detection-free replays for append/new_cluster; replace could re-run the same `_addExtremePeriods` replace branch on the incoming frame). Then transfer becomes exact for all three methods and the warning can go away.

---

## Why fix upstream

Both are the same shape of problem: extreme handling isn't *composable*. `append`/`new_cluster` transfer but aren't count-stable; `replace` is count-stable (`+0`) but doesn't transfer. Every tool that fans out or transfers clusterings has to pick a method that's wrong on one axis, or reimplement extreme detection to patch the gap. A predictable-count option (Problem 1) plus transferable `replace` (Problem 2) would make extremes usable as a building block.

---

## Reproduction

```python
"""tsam 3.4.1, no other deps beyond numpy/pandas."""
import warnings; warnings.filterwarnings("ignore", category=FutureWarning)
import numpy as np, pandas as pd, tsam

def make_df(seed, cols=("solar", "wind", "price")):
    r = np.random.default_rng(seed)
    idx = pd.date_range("2020-01-01", periods=30 * 24, freq="h")
    return pd.DataFrame({c: r.random(30 * 24) for c in cols}, index=idx)

# ---- Problem 1: number of appended extreme clusters is data-dependent ----
cfg = tsam.ExtremeConfig(method="append",
                         max_value=["solar", "wind", "price"],
                         min_value=["solar", "wind", "price"])
print("Problem 1 — n_clusters per dataset (n_clusters=8, 6 extreme criteria):")
for seed in range(6):
    res = tsam.aggregate(make_df(seed), 8, extremes=cfg)
    print(f"  seed={seed}: n_clusters={res.clustering.n_clusters}")

# ---- Problem 2: 'replace' silently drops its peak injection on apply() ----
print("\nProblem 2 — apply() a clustering back onto the SAME data:")
for method in ("append", "new_cluster", "replace"):
    df = make_df(0, cols=("solar", "wind"))
    ec = tsam.ExtremeConfig(method=method, max_value=["solar", "wind"],
                            min_value=["solar", "wind"])
    direct = tsam.aggregate(df, 8, extremes=ec)
    applied = direct.clustering.apply(df)
    reps_d = direct.cluster_representatives.values
    reps_a = applied.cluster_representatives.values
    maxdiff = np.abs(reps_d - reps_a).max()
    print(f"  method={method:11s} reps identical: {np.allclose(reps_d, reps_a)!s:5s}  "
          f"max|Δrep|={maxdiff:.3f}  "
          f"rmse {direct.accuracy.weighted_rmse:.4f} -> {applied.accuracy.weighted_rmse:.4f}")
```

Output:

```
Problem 1 — n_clusters per dataset (n_clusters=8, 6 extreme criteria):
  seed=0: n_clusters=10
  seed=1: n_clusters=13
  seed=2: n_clusters=11
  seed=3: n_clusters=10
  seed=4: n_clusters=12
  seed=5: n_clusters=12

Problem 2 — apply() a clustering back onto the SAME data:
  method=append      reps identical: True   max|Δrep|=0.000  rmse 0.2931 -> 0.2931
  method=new_cluster reps identical: True   max|Δrep|=0.000  rmse 0.2909 -> 0.2909
  method=replace     reps identical: False  max|Δrep|=0.966  rmse 0.3104 -> 0.3094
```

FBumann:
Fixes both problems in #410 — extreme-period handling that is neither count-stable nor transferable, which blocks building on top of `aggregate()`/`apply()`.

## Problem 1 — count stability

`append`/`new_cluster` return `n_clusters + k`, where `k` is *data-dependent*: criteria that resolve to the same period, or an extreme that already coincides with a cluster center, get deduplicated. So the same config over different datasets can yield different counts, and fanned-out results can't be stacked/aligned.

**Fix (opt-in):** `ExtremeConfig(preserve_n_clusters=True)` detects the distinct extremes up front, **excludes them from clustering**, clusters the rest into `n_clusters − D`, and adds the `D` extremes back — so the total is **exactly `n_clusters`**, independent of the data.

```python
cfg = ExtremeConfig(method="append", preserve_n_clusters=True,
                    max_value=["solar", "wind"], min_value=["solar", "wind"])
# every dataset -> exactly n_clusters typical periods
```

- Applies to `append`/`new_cluster`; **no-op with a warning** for `replace` (already count-stable).
- Requires `n_clusters > D` (clear `ValueError` otherwise).
- Trade-off: fewer *regular* clusters, so ordinary periods are represented a little more coarsely — the price of a config-determined count. Column-mean rescaling and extreme preservation are unaffected.
- `new_cluster`'s reassignment can't drop a regular cluster below non-empty (representatives are pinned), so the total holds unconditionally.

## Problem 2 — transferable `replace`

`replace` built a hybrid center with no addressable period index, so `apply()` silently dropped its peak injection and fell back to the plain medoid.

**Fix:** store each `(cluster, column, period)` injection on `ClusteringResult` and **replay** it in `apply()`/JSON, overwriting the same column of the same cluster using the *new* data at the same period index (consistent with how `append`/`new_cluster` replay stored indices). All three methods now round-trip exactly. The old warning now only fires for clusterings produced before this support (missing stored replacements).

## Compatibility

Both behaviors are **opt-in / additive** — default outputs and golden CSVs are unchanged (no regeneration).

## Testing

- New `test/test_extreme_composability.py`: count invariance across datasets/methods/weights/segments/duration-curves, collision case, `new_cluster` non-emptying, an adversarial "extreme is also a natural center" case, mean preservation, the `n_clusters − D < 1` error, exact `replace` round-trips (apply + JSON) and warning behavior, plus a **50-case fuzz** over random data/methods/criteria/budgets.
- Enabled `replace` in the e2e transfer suite (removed the old exclusion).
- Full suite green (535 passed), golden regression intact.

## Docs

New "Extreme periods" section in `docs/background/math.md` (criteria, methods, count stability, transfer); `ExtremeConfig` / `add_extreme_periods` docstrings document `preserve_n_clusters` and the count behavior.

Closes #410


FBumann:
Blocked by #479

PR #418, opened 2026-07-23
feat(clustering): KMedoids method config forwarding solver options (#365)

### LINKED ISSUE #365: KMedoids: timelimit and threads are stored but never passed to the solver
### Summary

`KMedoids` accepts `timelimit` (default `100`) and `threads` (default `7`) and stores them on the instance, but **neither is ever passed to the solver**. The exact k-medoids MILP therefore runs with no enforced time limit and the configured thread count is ignored.

### Where

`src/tsam/algorithms/k_medoids_exact.py`

The constructor stores both values:

```python
def __init__(self, n_clusters=8, distance_metric="euclidean",
             timelimit=100, threads=7, solver="highs"):
    ...
    self.timelimit = timelimit
    self.threads = threads
```

But `_solve_given_pyomo_model` solves without using either:

```python
def _solve_given_pyomo_model(M, solver="highs"):
    if solver == "highs":
        solver_instance = appsi.solvers.Highs()
    else:
        solver_instance = opt.SolverFactory(solver)
    _results = solver_instance.solve(M)   # no time limit, no thread count
    ...
```

`self.timelimit` / `self.threads` are never read after assignment, and `_solve_given_pyomo_model` does not receive them.

### Impact

- The docstring advertises a time limit (`:param timelimit: Specify the time limit of the solver.`) that is not enforced. On large period counts the model has `n²` binary variables and can run unbounded.
- The `threads` setting has no effect.

### Suggested fix

Thread `timelimit`/`threads` through to the solver, e.g. for the appsi HiGHS path set `solver_instance.config.time_limit` / the HiGHS `threads` option, and for the `SolverFactory` path pass solver options / `timelimit=` on `.solve(...)`. Add a regression test that a tiny model respects a very small time limit (or at least that the options are forwarded).

Found while reviewing the algorithm implementations for the v4 documentation rework.

FBumann:
**Draft.** Rebased onto `develop` on 2026-07-31 (it was based on a pre-#234 `develop-v4`). #454 stacks on this one, adding a `KMeans` config to the same union.

<details><summary>What the rebase had to resolve</summary>

`develop` had since typed and Google-docstring'd `k_medoids_exact.py` and extracted `DEFAULT_REPRESENTATION` into `config.py`, so 9 conflict markers across 4 files. Resolved by keeping develop's modernised form and layering this PR's semantics on top. Three defects the auto-merge introduced that the markers did *not* flag, all fixed:

- `config.py`: develop's `use_duration_curves` validation warning was stranded after a `return` inside the new `solver` property — dead code, and one test failed on it. Moved back into `__init__`.
- `k_medoids_exact.py`: the `KMedoids` → `ExactKMedoids` rename missed a return annotation.
- A new mypy error — `DEFAULT_REPRESENTATION[cluster.method]` no longer type-checks now that `method` is a union; changed to `cluster.method_name`.

Full suite **709 passed**; mypy back to develop's 4 pre-existing errors; ruff clean.
</details>

---

## Summary

Fixes #365. The exact k-medoids MILP stored a `timelimit`/`threads` that were **never passed to the solver**, so the advertised time limit was silently ignored. They were also unreachable from `aggregate()` — only `solver` was.

This adds a **`KMedoids` method-config object**, modelled on `Distribution` for representations:

```python
from tsam import aggregate, ClusterConfig, KMedoids

aggregate(df, n_clusters=8,
          cluster=ClusterConfig(method=KMedoids(solver="highs",
                                                options={"time_limit": 300})))
```

Rather than enumerating individual knobs, **`options` is a plain dict forwarded verbatim to the solver** (native option names, e.g. `{"time_limit": 300}` for HiGHS). This keeps tsam out of the business of tracking per-solver knobs and makes every solver option reachable.

## Design

- `ClusterMethod` is now the union `ClusterMethodName | KMedoids` (mirrors `Representation = RepresentationMethod | Distribution | MinMaxMean`). Future methods can add their own config classes to the union.
- The bare string `method="kmedoids"` still works (maps to `KMedoids()` defaults).
- `method` is stored **raw** and dispatched by type; nested-dict serialization (`{"method": {"type": "kmedoids", ...}}`) round-trips + pickles identically. Legacy flat `{"method": "kmedoids", "solver": ...}` JSON still loads.
- `ClusterConfig(solver=...)` is **deprecated** and forwarded into the object with a warning.
- `KMedoids` is hashable regardless of `options` (dict excluded from `__hash__`, kept in `__eq__`), so `ClusterConfig` stays hashable.
- Internal estimator renamed `KMedoids` → `ExactKMedoids` to free the public name (single internal caller).

## `threads` was deliberately dropped as a dedicated knob

Forwarding `threads` to the appsi HiGHS interface caused **nondeterministic** "no feasible solution" failures on small MILPs (one process fails on `threads=1/2`, another passes 4×). It had no demonstrated demand. Users who really want it can still pass `options={"threads": N}` at their own risk.

## Behavior note

There is **no** default time limit (matching historical behavior — the old default was never applied). Pass `options={"time_limit": ...}` to bound runtime; otherwise the solver's own defaults apply.

## Tests

New `test/test_kmedoids_config.py` (14 tests): options forwarding to the solver (the #365 regression), string/object equivalence, hashability, deprecation warning, serialization round-trips, legacy-dict loading. Full suite: **469 passed, 0 failed**; pre-commit clean (ruff, mypy, API-ref coverage).


julian-belina:
@FBumann, is this ready to be merged? It's marked as a draft and should be merged before https://github.com/FZJ-IEK3-VSA/tsam/pull/454, which is not a draft.

julian-belina:
@FBumann, quick update. Is this ready for review? There are many open PRs and it's no longer clear which ones are drafts and which ones are ready for review.

FBumann:
All of them are draft or not?
I would need to take another look to verify they are ready. But you can also take on then if you have the time

julian-belina (approved):

PR #439, opened 2026-07-29
fix(algorithms): break representative-selection ties deterministically

FBumann:
> [!IMPORTANT]
> **Superseded — see the correction below.** The original note claimed six
> commits across #434–#439 needed `BREAKING CHANGE:` footers. Measuring each
> against v3.4.2 cut that to **two**, and the two commits that genuinely needed
> footers were given them during the squash merges.
>
> Removed from the list, having been measured identical to v3.4.2:
> - *period-sum features weighted* and *period sums kept out of duration-curve
>   clustering* — both **restore** v3 behaviour, so they are `fix` and change
>   nothing for a v3 user.
> - *`MinMaxMean` raising `ValueError`* and *cluster/segment representation
>   split* — real, and both now carry footers on `develop-v4` (`5d2dedf`,
>   `21b7831`).
>
> Still lacking a footer, both real behaviour changes:
> - `5a8d49b` deterministic tie-breaking — moves one golden (max abs 21.75)
> - `fe80a46` `apply()` replaying the stored `ClusterConfig` — changes
>   transferred results when `scale_by_column_means` is set
>
> *duration-curve center indices* sits inside `5d2dedf`, which already carries a
> footer; it affects `ClusteringResult.cluster_centers` and transfers only, not
> `aggregate()` output.

Stacked on #438 → #437 → #436 → #435 → #434. Closes the last `xfail` from #435.

## What `segmentation_medoid` actually was

Not a segmentation change. The segment **durations are byte-identical** to v3.4.2, so the agglomerative segmentation agrees completely; only the medoid *pick* inside one segment differed — **2 cells of 256**.

That segment holds exactly **two timesteps**, so both members are equidistant and the medoid is mathematically undefined. Instrumenting both versions shows the pick decided by one ULP:

```
v4:  seg-cluster 6: sum0-sum1 = -5.4817e-16  → argmin=0   (Load=498.98)
v3:  seg-cluster 6: sum0-sum1 = +5.4817e-16  → argmin=1   (Load=481.02)
```

Same magnitude, opposite sign. The sign comes from **column order** — v3 sorted columns alphabetically, v4 preserves input order, so `euclidean_distances` accumulates the squared differences in a different order. The already-documented column-order change turns out to be able to move *values*, via this tie.

## How exposed this was

Across the golden matrix on `testdata`: **464 medoid selections, 23 exactly tied, 2 near-tied.** The exact ties are stable (`argmin` returns 0); the near-ties are the ones that flip with BLAS build, array layout or column order. The same code on the same data could pick a different representative on a different machine.

## The fix

`algorithms/concurrency.py` had already solved this locally with a private `_deterministic_argmin` — rounding before comparing so noise-level differences collapse to a true tie, which `argmin` resolves to the lowest index. It just never reached the representations.

That helper is promoted to `algorithms/selection.py` and now backs all four selection sites:

| Site | Was |
|---|---|
| `medoid_representation` | `np.argmin` |
| `maxoid_representation` | `np.argmax` |
| `cluster_sorted_periods` (duration-curve centers) | `np.argmin` — same pattern, same exposure |
| `concurrency.compute_ordering` | already had it, now shared |

It lives under `algorithms/` rather than `commons` so the kernels stay self-contained, and in its own module so `concurrency` can use it without closing the `representations → duration_representation → concurrency` import loop.

## The one regenerated golden

`segmentation_medoid/testdata` is the only golden in the matrix **not** produced by v3.4.2, and `_golden_cases.py` says why at the case: for that segment there is no v3 answer to reproduce, only a v3 coin flip. The regenerated file records the deterministic choice — which is what v3 would also have produced had it been deterministic.

**Every other golden is unchanged.** The fix moves nothing that was not already undefined; I ran the full matrix before regenerating and 143 of 144 passed untouched.

## Tests

`test_representation_tiebreak.py` pins the property directly:

- a two-member group resolves to the earlier member, for medoid and maxoid
- a one-ULP perturbation does not flip the pick
- reordering the input columns changes neither the assignment nor the selected centers

## Verification

- `pytest test/`: 687 passed, 145 skipped, 5 xfailed, 1 xpassed
- `ruff check` / `ruff format --check`: clean
- `mkdocs build --strict`: clean


julian-belina (approved):

julian-belina:
@FBumann Thank you very much for your work. I have finished my review, and we can proceed with the V4 release.

julian-belina (approved):

PR #440, opened 2026-07-30
test(golden): bring the v3 line up to the v4 golden matrix (116 -> 144)

FBumann:
Ports `test/_golden_cases.py` from `develop-v4` onto the v3 line and generates
every golden from it, taking `test/data/golden/` from **116 files (62
configurations) to 144 (90)**.

## Why

The v4 branch widened the matrix to cover options no configuration exercised at
all — the string forms of `medoid` and `minmax_mean`, every
`SegmentConfig.representation` other than `"mean"`, any `period_duration` other
than 24, an explicit `temporal_resolution`, extremes on
`averaging`/`kmedoids`/`kmaxoids`, duration curves on any method other than
`hierarchical`/`kmeans`, `include_period_sums` in combination with anything,
`normalize_column_means` without weights, and any `n_clusters` other than 8.

The v3 line is still the released line, and none of that was covered here.

## What changed to make the matrix run

Two adjustments, both naming only:

- the shared dataset lives under `docs/notebooks/` here, `docs/data/` on v4
- the option v4 spells `scale_by_column_means` is `normalize_column_means` here

The input CSVs are byte-identical across the two branches, so nothing about the
inputs differs.

### 1. Regenerating on this branch's own code — the diff you see here

Regenerating every golden reproduced **112 of the 116** existing files
byte-for-byte. Four differ, and only in a way worth naming exactly:

| File | Cells differing | Column | Delta |
|---|---|---|---|
| `hierarchical_mean/testdata.csv` | 52 / 35 040 | `Load` only | `1e-08` |
| `hierarchical_mean/with_zero_column.csv` | 52 / 43 800 | `Load` only | `1e-08` |
| `kmeans_weighted/testdata.csv` | 192 / 35 040 | `Load` only | `1e-08` |
| `kmeans_weighted_segmentation/testdata.csv` | 160 / 35 040 | `Load` only | `1e-08` |

Across all four there is a **single distinct delta** — exactly `1e-08`, the last
digit of `round(8)`. For example `489.90639802 -> 489.90639803`. `Load` carries
the largest magnitudes of the four attributes, so it has the coarsest floating-
point spacing and is the column that lands on a rounding boundary.

`develop`'s bytes are kept for these four, matching
`chore: restore develop's bytes for 4 sub-tolerance goldens`. So the diff here
is 28 new goldens plus the ported case file — **+232 059 / −6** — and no
reviewer has to wonder whether four numerical files quietly moved.

That the committed v3 goldens regenerate byte-identically apart from four
rounding boundaries is itself a useful result: they were already faithful to
this branch's code.

### 2. Against `develop-v4` — the actual v3 to v4 delta

Because this branch is "v3 code, v4 matrix", diffing its goldens against
`develop-v4`'s measures the change between the lines directly rather than
arguing it from the code. Across all 144, **five** files differ: the same four
`1e-08` rounding boundaries above, plus one real difference —

| File | Max abs |
|---|---|
| `segmentation_medoid/testdata.csv` | **21.75** |

That configuration has a segment of exactly two timesteps. Both members are
therefore equidistant and the medoid is mathematically undefined; v3 resolves it
with a bare `argmin`, which floating-point noise decides, while v4 breaks the
tie deterministically.

**So v3.4.x and v4 agree numerically across all 90 configurations except one
undefined tie.** That is a stronger statement than the migration guide currently
makes, and it is now reproducible by anyone rather than asserted.

## Note on `segmentation_medoid`

Its golden here holds v3's value, which is correct for this line — v3 still has
the non-deterministic `argmin`. It is the one intentional difference between the
two branches' goldens. Backporting the tie-break would make v3's medoid
selection reproducible across machines, and is worth considering separately.

## Verification

- `pytest test/`: 412 passed, 160 skipped
- `pytest test/test_golden_regression.py`: 144 passed — every generated golden
  reproduces on this branch's code


FBumann:
@julian-belina This ports all the new tests to the last tsam release. Please approve, so the diff of develop-v4 shows them all as equal!

julian-belina (approved):

PR #444, opened 2026-07-31
perf: vectorize new_cluster extreme-period reassignment

FBumann:
**Draft.** Stacked on #443. One change, measured against it.

`add_extreme_periods` decided, for every period independently, whether some extreme period was closer than the cluster center it had been assigned to. Each decision pulled the period out with `profiles_df.iloc[i].values` and summed squared distances with the builtin `sum()`, so the work scaled with periods x extremes in interpreted Python — 1095 periods over three years, each compared against every extreme.

Both distance sets are now computed once by numpy broadcasting. `argmin` with a strict comparison against the own-center distance preserves the first-match-on-ties semantics of the original scan.

## Affects

Only `ExtremeConfig(method="new_cluster")`. **Not** `"append"`, **not** `"replace"` — those take different branches and are untouched.

## Isolated effect

|---|---:|---:|---:|
| `test_extremes_3y` (3 y, new_cluster) | 56.89 ms | 31.68 ms | **−44.3%** |
| `test_feature[extremes]` (1 y, new_cluster) | 19.54 ms | 11.34 ms | **−42.0%** |
| `test_fine_extremes` (append — control) | 88.60 ms | 88.37 ms | −0.3% |

The control is the point: `append` never reaches this code and does not move.

## Workflow effect

**None measurable — and that is expected.** No workflow in `test_workflows.py` configures `new_cluster`; the one that uses extremes at all (`test_workflow_energy_system_model[mean_extremes]`) uses `append`. All twelve workflow measurements sat inside ±5%.

A workflow that does use `new_cluster` would gain in proportion to how much of its time is extreme reassignment — which the isolated 3-year case puts at roughly 44% of that stage.

## Measurement protocol

## Equivalence

Identical cluster assignments against the previous scalar loop over 200 randomised trials (varying period counts, columns, timesteps, cluster counts, and which extremes are configured). Full test suite green.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -76,3 +76,4 @@ docs/**/clustering.json
 
 # uv resolves this on demand; do not track it
 uv.lock
+benchmarks/reports/
--- a/benchmarks/plot_report.py
+++ b/benchmarks/plot_report.py
@@ -1,0 +1,212 @@
+#!/usr/bin/env python
+"""Build a standalone HTML performance report from pytest-benchmark runs.
+
+Reads two (or more) benchmark JSON files, pairs them by benchmark id, and
+writes one self-contained page: scaling curves on log-log axes plus a table
+attributing the change per benchmark. The output has plotly inlined, so it
+opens from disk with no network.
+
+The dims the suite stamps (`n_columns`, `n_clusters`, `method`, ...) are what
+make this possible: `pytest_benchmem.load_long_df` turns each run into a tidy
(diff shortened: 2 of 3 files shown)


FBumann:

- `test_extremes_3y` 56.89 → 31.68 ms (−44.3%), `test_feature[extremes]` 19.54 → 11.34 ms (−42.0%); `append` control unmoved.
- No workflow effect: no workflow in the suite configures `new_cluster`.
Branch `perf/extremes-new-cluster` is retained.

PR #445, opened 2026-07-31
perf: cache decision arrays in the k-maxoids inner loop

FBumann:
**Draft.** Stacked on #444. One change, measured against it.

The k-maxoids inner loop recomputed each sample's nearest maxoid, its own value and the maxoid's value on every pass, though a pass only changes them when it performs a replacement — which it does in roughly 1.5% of iterations. Those decision inputs are now cached and refreshed only when a replacement modifies `M`. The loop structure and the per-sample decision are unchanged.

## Affects

Only `ClusterConfig(method="kmaxoids")`. No other clustering method reaches this code.

## Isolated effect

|---|---:|---:|---:|
| `test_method[kmaxoids]` (8760 x 4, 8 clusters) | 1311 ms | 227 ms | **−82.7%** (5.8x) |

## Workflow effect

**None measurable.** No workflow in the suite selects `kmaxoids` — the setups exercised are `hierarchical_medoid`, `kmeans_mean`, `duration_segments` and `mean_extremes`. All twelve workflow measurements sat inside ±5%.

The isolated number is nevertheless the honest one for anyone who does select it: k-maxoids is by a wide margin the most expensive method in the suite, and this removes most of that cost. For scale, on the same input `hierarchical` takes ~11 ms — so this closes the gap from 120x to 20x rather than eliminating it.

## Measurement protocol

## Equivalence

Bit-identical over 200 seeded randomised trials: cluster centers and labels hash-equal before and after, across varying sample counts, feature counts and cluster counts. Full test suite green.

Independently of this PR, worth knowing: `kmaxoids` seeds from an unseeded global RNG, so two `aggregate()` calls on the *same* tree can differ (two runs reassigned 351 of 365 periods). That is why equivalence here is asserted at the algorithm level under a fixed seed rather than end to end.

### DIFF
--- a/src/tsam/algorithms/k_maxoids.py
+++ b/src/tsam/algorithms/k_maxoids.py
@@ -97,6 +97,42 @@ class KMaxoids(BaseEstimator, ClusterMixin, TransformerMixin):
 
         return X
 
+    @staticmethod
+    def _run_passes(X: np.ndarray, M: np.ndarray, n_passes: int) -> None:
+        """Run the sequential maxoid-replacement passes, updating ``M`` in place.
+
+        Same loop and per-sample decision as the naive implementation, but the
+        three decision inputs (nearest maxoid, sample value, maxoid value) are
+        read from arrays that only need recomputing when a replacement modifies
+        ``M`` — a rare event. The recomputation uses the same
+        ``np.sum((A - b) ** 2, axis=1)`` reduction as the naive per-sample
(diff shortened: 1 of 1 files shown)


FBumann:

- `test_method[kmaxoids]` 1311 → 227 ms (−82.7%), bit-identical over 200 seeded trials.
- No workflow effect: no workflow selects `kmaxoids`.
Branch `perf/kmaxoids-cache` is retained.

PR #446, opened 2026-07-31
perf: numpy reshapes for the regular period grid

FBumann:
**Draft.** Stacked on #445. One change, measured against it.

`unstack_to_periods` built a two-level MultiIndex from two Python lists, one entry per timestep, then called `DataFrame.unstack` to pivot it; `reconstruct` did the inverse with `stack`. Both are general enough for ragged input, and the input here is never ragged — a period grid is rectangular by construction, which makes the pivot a `reshape` plus a `transpose`.

The fast path applies only when every column shares one plain numpy dtype. Its labels come from unstacking a single period through pandas, so they are what `unstack` would have produced rather than a hand-built index that happens to look right. Mixed-dtype frames keep the pandas path.

## Affects

**Every `aggregate()` call**, and the public `tsam.unstack_to_periods()`. This is on the main path, not behind a config option. Mixed-dtype input falls back and is unaffected.

## Isolated effect

Six repetitions, per-rep spread under 1%:

|---|---:|---:|---:|
| `test_resolution_15min` (96 steps/period) | 26.49 ms | 11.59 ms | **−56.2%** |
| `test_scale_columns[4]` | 10.43 ms | 6.39 ms | **−38.7%** |
| `test_scale_columns[12]` | 14.92 ms | 10.86 ms | **−27.2%** |
| `test_scale_columns[48]` | 35.20 ms | 31.13 ms | **−11.6%** |
| `test_scale_timesteps[1 y]` | 35.38 ms | 31.17 ms | **−11.9%** |
| `test_scale_timesteps[3 y]` | 179.23 ms | 170.96 ms | **−4.6%** |

The benefit **shrinks as the input widens** — at 400 columns clustering dominates and the reshape is a small share. 15-min resolution gains most because it has 96 timesteps per period rather than 24, so the reshape it replaces was four times larger.

## Workflow effect

Largest on workflows that call `aggregate()` many times, where the fixed reshape cost is paid per call rather than amortised:

|---|---:|---:|---:|
| `tuning_sweep[hierarchical_medoid]` | 222.5 ms | 158.4 ms | **−28.8%** |
| `scenario_batch[hierarchical_medoid]` | 189.5 ms | 154.3 ms | **−18.6%** |
| `tuning_sweep[duration_segments]` | 450.2 ms | 395.8 ms | **−12.1%** |

Single-call workflows moved less than the ±5% noise floor.

<details><summary>Workflow code (<code>benchmarks/test_workflows.py</code>)</summary>

```python
@pytest.mark.parametrize("setup", ["hierarchical_medoid", "duration_segments"])
def test_workflow_tuning_sweep(benchmark, setup):
    """Search for the smallest typical-period count that still fits the data.

    Twelve candidate cluster counts, each scored by its duration-curve error —
    the loop ``tsam.tuning`` runs internally, written out so it behaves
    identically on every version being compared.
    """
    measurements = pd.read_csv(TESTDATA_CSV, index_col=0, parse_dates=True)
    candidates = [4, 6, 8, 10, 12, 16, 20, 24, 30, 36, 48, 60]
    kwargs = _setup_kwargs(setup, [])
    benchmark.extra_info["n_configs"] = len(candidates)

    def run():
        scored = []
        for n_clusters in candidates:
            result = aggregate(measurements, n_clusters, **kwargs)
            scored.append((n_clusters, float(result.accuracy.rmse_duration.mean())))
        return min(scored, key=lambda candidate: candidate[1])

```
</details>

## Measurement protocol

## Equivalence

Golden regression suite green; full test suite green. This changes floating-point reduction order, so the reconstructed series is numerically equal rather than bit-identical — at most ~1e-13 absolute on values of order 1e3. No cluster assignment moves.

### DIFF
--- a/src/tsam/pipeline/accuracy.py
+++ b/src/tsam/pipeline/accuracy.py
@@ -55,20 +55,20 @@ def reconstruct(
     """
     # Unstack once, then use vectorized indexing to select periods by cluster order
     typical_unstacked = typical_periods.unstack()
-    reconstructed = typical_unstacked.loc[list(cluster_order)].values
+    reconstructed = typical_unstacked.loc[list(cluster_order)].to_numpy()
 
-    # Back in matrix form
-    clustered_data_df = pd.DataFrame(
-        reconstructed,
-        columns=period_profiles.column_index,
-        index=period_profiles.profiles_dataframe.index,
+    # Rows are period vectors in (column, timestep) layout, so stacking back to
+    # the flat time series is a plain numpy reshape; trim to original length.
+    n_columns = period_profiles.n_columns
+    n_timesteps = period_profiles.n_timesteps_per_period
+    flat = (
+        reconstructed.reshape(len(cluster_order), n_columns, n_timesteps)
+        .transpose(0, 2, 1)
+        .reshape(-1, n_columns)
     )
-    clustered_data_df = clustered_data_df.stack(future_stack=True, level="TimeStep")  # type: ignore[assignment]
 
-    # Trim to original data length
-    original_len = len(original_data)
     normalized_predicted = pd.DataFrame(
-        clustered_data_df.values[:original_len],
+        flat[: len(original_data)],
         index=original_data.index,
         columns=original_data.columns,
     )
--- a/src/tsam/pipeline/periods.py
+++ b/src/tsam/pipeline/periods.py
@@ -2,8 +2,8 @@
 
 from __future__ import annotations
 
-import copy
 import warnings
+from typing import cast
 
 import numpy as np
 import pandas as pd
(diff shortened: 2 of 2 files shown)


FBumann:

- Six repetitions, spread under 1%: `test_resolution_15min` −56.2%, `test_scale_columns[4/12/48]` −38.7/−27.2/−11.6%, `test_scale_timesteps[1/2/3 y]` −11.9/−6.8/−4.6%.
- Workflow: `tuning_sweep` −28.8%, `scenario_batch` −18.6%.
- Note the benefit shrinks as inputs widen, as clustering takes over.
Branch `perf/period-grid-reshape` is retained.

PR #448, opened 2026-07-31
perf: reduce section means with np.add.reduceat

FBumann:
**Draft.** Stacked on #447. One change, measured against it.

The remaining loop in global duration representation averaged each section with its own `np.mean` call — `n_clusters x timesteps` of them, each over a handful of values, where per-call overhead dominates the arithmetic. One `np.add.reduceat` over the section boundaries replaces all of them.

Deliberately separate from #447: a different transformation on the same function, and folding them together would have made neither measurable on its own.

## Affects

The same three entry points as #447 — the global branch of `duration_representation` only.

## Isolated effect

|---|---:|---:|---:|
| `test_representation_global_scale[48]` | 58.10 ms | 48.59 ms | **−16.4%** |
| `test_representation[distribution_global]` | 8.13 ms | 7.16 ms | **−11.9%** |
| `test_representation_global_scale[4]` | 8.26 ms | 7.42 ms | **−10.2%** |
| `test_representation[mean/medoid/maxoid/minmax_mean]` | — | — | within ±5% |

## Workflow effect

**At or below the noise floor.** The largest was `scenario_batch[hierarchical_medoid]` at −5.8%; everything else sat inside ±5%. This is a real but small refinement on top of #447, and the honest statement is that it is not visible in a whole job.

## Measurement protocol

## Equivalence

Bit-exact: `reduceat` sums the same values in the same order as the per-section `mean` it replaces. Golden regression suite green; full test suite green.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -76,3 +76,4 @@ docs/**/clustering.json
 
 # uv resolves this on demand; do not track it
 uv.lock
+benchmarks/reports/
--- a/benchmarks/plot_report.py
+++ b/benchmarks/plot_report.py
@@ -1,0 +1,212 @@
+#!/usr/bin/env python
+"""Build a standalone HTML performance report from pytest-benchmark runs.
+
+Reads two (or more) benchmark JSON files, pairs them by benchmark id, and
+writes one self-contained page: scaling curves on log-log axes plus a table
+attributing the change per benchmark. The output has plotly inlined, so it
+opens from disk with no network.
+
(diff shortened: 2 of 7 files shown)


FBumann:

- `test_representation_global_scale[48]` −16.4%, `[4]` −10.2% on top of #447.
- No workflow effect above the ±5% noise floor.
This is a refinement of #447 and is best re-filed directly after it. Branch `perf/duration-section-means` is retained.

PR #449, opened 2026-07-31
perf: compute accuracy metrics on the whole frame

FBumann:
**Draft.** Targets `develop` directly: one file, independent of every other PR in the performance stack. Measured against the suite in #443.

`compute_accuracy` looped over columns, calling scikit-learn's `mean_squared_error` and `mean_absolute_error` once per column and sorting each column twice with `sort_values` to build its duration curve. Every one of those calls re-validates its inputs, and the validation cost more than the arithmetic. The metrics are now computed on the whole frame in numpy, sorting once along an axis.

## Affects

`AggregationResult.accuracy` — every field (`rmse`, `mae`, `rmse_duration` and the weighted variants) — plus `result.to_dict()`, which serialises them.

`accuracy` is a lazy `cached_property`. **No `aggregate()`-only benchmark ever charged for this**, at any input size; it looked free until the suite in #443 grew a benchmark that materialises it. Anyone reporting error metrics, or sweeping configurations scored by them, pays it on every call.

## Isolated effect

|---|---:|---:|---:|
| `test_accuracy[48]` | 53.20 ms | 13.85 ms | **−74.0%** |
| `test_accuracy[4]` | 4.71 ms | 1.31 ms | **−72.2%** |

At 400 columns the same path goes 440 → 114 ms, so the ratio holds as the frame widens.

## Workflow effect

|---|---:|---:|---:|
| `accuracy_report[hierarchical_medoid]` | 89.66 ms | 49.95 ms | **−44.3%** |
| `scenario_batch[hierarchical_medoid]` | 181.80 ms | 125.50 ms | **−31.0%** |
| `tuning_sweep[hierarchical_medoid]` | 195.09 ms | 161.97 ms | **−17.0%** |
| `tuning_sweep[duration_segments]` | 440.48 ms | 394.20 ms | **−10.5%** |

Workflows that never read `accuracy` sat inside the noise floor, as they should.

<details><summary>Workflow code (<code>benchmarks/test_workflows.py</code>, added in #443)</summary>

```python
@pytest.mark.parametrize("setup", ["hierarchical_medoid", "kmeans_mean"])
def test_workflow_accuracy_report(benchmark, setup):
    """Aggregate a wide dataset and report how much accuracy it cost.

    What someone does when choosing between configurations: run one, look at
    the per-attribute error. ``accuracy`` is lazy, so this is the only
    workflow that pays for it on a single call.
    """
    profiles = pd.read_csv(WIDE_CSV, index_col=0, parse_dates=True)
    profiles = pd.concat(
        [profiles.add_suffix(f"_{i}") for i in range(12)], axis=1
    ).iloc[:, :48]
    kwargs = _setup_kwargs(setup, [])

    def run():
        result = aggregate(profiles, 8, **kwargs)
        accuracy = result.accuracy
        return accuracy.rmse, accuracy.mae, accuracy.rmse_duration

    benchmark.pedantic(run, **LIGHT)


@pytest.mark.parametrize("setup", ["hierarchical_medoid", "kmeans_mean"])
def test_workflow_scenario_batch(benchmark, setup):
    """Aggregate eight scenario years in one job, scoring each.

    Batch jobs multiply per-call overhead by the number of scenarios instead
    of amortizing it, so a fixed cost shows up here that a single large
    aggregation would hide.
    """
    profiles = pd.read_csv(WIDE_CSV, index_col=0, parse_dates=True)
    wide = pd.concat([profiles.add_suffix(f"_{i}") for i in range(16)], axis=1)
    scenarios = [wide.iloc[:, i * 8 : (i + 1) * 8] for i in range(8)]
    kwargs = _setup_kwargs(setup, [])
    benchmark.extra_info["n_scenarios"] = len(scenarios)

    def run():
        return [
            aggregate(scenario, 8, **kwargs).accuracy.rmse.mean()
            for scenario in scenarios
        ]

@pytest.mark.parametrize("setup", ["hierarchical_medoid", "duration_segments"])
def test_workflow_tuning_sweep(benchmark, setup):
    """Search for the smallest typical-period count that still fits the data.

    Twelve candidate cluster counts, each scored by its duration-curve error —
    the loop ``tsam.tuning`` runs internally, written out so it behaves
    identically on every version being compared.
    """
    measurements = pd.read_csv(TESTDATA_CSV, index_col=0, parse_dates=True)
    candidates = [4, 6, 8, 10, 12, 16, 20, 24, 30, 36, 48, 60]
    kwargs = _setup_kwargs(setup, [])
    benchmark.extra_info["n_configs"] = len(candidates)

    def run():
        scored = []
        for n_clusters in candidates:
            result = aggregate(measurements, n_clusters, **kwargs)
            scored.append((n_clusters, float(result.accuracy.rmse_duration.mean())))
        return min(scored, key=lambda candidate: candidate[1])

```
</details>

## Measurement protocol

Arms interleaved within each repetition; **minimum** across four runs, since contention only ever adds time. Noise floor about ±5%, so nothing smaller is claimed. Workflows whose setup is `kmeans_mean` are excluded from the quoted figures — multithreaded scikit-learn makes them swing up to ~20% run to run. macOS arm64, Python 3.12, numpy 2.5.1 / pandas 3.0.5 / scikit-learn 1.8.0 / scipy 1.17.0.

## Equivalence

Golden regression suite green; full test suite green (695 passed at this level). Reduction order changes, so metrics are numerically equal rather than bit-identical (~1e-16 relative).

Residual: roughly 75% of what remains is `np.sort` itself, which the duration metric requires.

## Reviewer note

+14/−21 in one file. No new tests — it relies on the existing coverage in `test_accuracyIndicators.py` and the ~10 other files asserting on `.rmse`/`.mae`. That coverage pins behaviour but was not written to catch a reduction-order change; a property test comparing old scalar against new vectorised output over randomised input would close that gap and I am happy to add one.


julian-belina (approved):
Makes sense, thanks for your contribution

Issue #432, opened 2026-07-27
Performance: ClusteringResult.disaggregate() is ~25x slower than a numpy gather


FBumann:
## Summary

`ClusteringResult.disaggregate()` is ~25x slower than necessary. `_expand_periods` implements what is a plain gather along the cluster axis as `unstack()` → `.loc[assignments]` → `stack()`, and `_validate_disaggregate_input` adds a further ~33% to every call.

This is a performance issue only — the output is correct. It matters when disaggregating many variables against the same clustering (e.g. expanding an optimization result), where the cost is paid once per variable.

## Reproduction

Pure pandas in, pandas out — no wrapper libraries involved. 36 clusters x 24 timesteps, 365 original periods, 20 columns, best of 30 runs, tsam 3.4.0 / Python 3.13:

```python
import numpy as np, pandas as pd, time
from tsam.config import ClusteringResult, _expand_periods

N_CLUSTERS, L, N_ORIG, N_COLS = 36, 24, 365, 20
rng = np.random.default_rng(0)
assignments = tuple(int(x) for x in rng.integers(0, N_CLUSTERS, N_ORIG))
mi = pd.MultiIndex.from_product([range(N_CLUSTERS), range(L)], names=["cluster", "timestep"])
df = pd.DataFrame(rng.random((N_CLUSTERS * L, N_COLS)), index=mi, columns=range(N_COLS))

def numpy_expand(df, assignments):
    """Same operation: reshape -> gather -> reshape."""
    n_clusters = df.index.get_level_values(0).nunique()
    n_t = len(df) // n_clusters
    labels = df.index.get_level_values(0).unique().to_numpy()
    pos = {int(v): i for i, v in enumerate(labels)}
    idx = np.fromiter((pos[a] for a in assignments), dtype=np.intp, count=len(assignments))
    cube = df.to_numpy().reshape(n_clusters, n_t, df.shape[1])
    out = cube[idx].reshape(len(assignments) * n_t, df.shape[1])
    return pd.DataFrame(out, index=pd.RangeIndex(len(out)), columns=df.columns)

ref, got = _expand_periods(df, assignments), numpy_expand(df, assignments)
assert ref.shape == got.shape and np.allclose(ref.to_numpy(), got.to_numpy())  # passes
```

Results:

```
tsam _expand_periods (pandas):    2.44 ms
numpy equivalent            :    0.10 ms
-> speedup                  : 25.0x

[...]

PR #450, opened 2026-07-31
perf: expand disaggregate() with numpy gathers

FBumann:
**Draft.** Stacked on #449. One change, measured against it. Addresses #432.

`ClusteringResult.disaggregate()` expanded typical-period data back onto the full time index by round-tripping through `unstack`/`stack`, rebuilding a MultiIndex on every call. Expansion is a gather: each original timestep already knows which typical period and timestep it maps to, so the result is one fancy-index into an array. `_expand_periods` and the segment expansion now do that directly, and input validation stops rebuilding frames to check shapes.

## Affects

`ClusteringResult.disaggregate()` and `AggregationResult.disaggregate()` — the caller-invoked expansion of external results. **Not** the reconstruction of the input, which is eager inside `aggregate()` and takes a different path.

Like `accuracy` in #449, this is code an `aggregate()`-only benchmark never touches — which is why the headroom was there to find.

## Isolated effect

|---|---:|---:|---:|
| `test_disaggregate[wide_segmented]` (36 clusters x 20 col, 12 seg) | 8.11 ms | 0.48 ms | **−94.0%** (17x) |
| `test_disaggregate[wide]` (36 clusters x 20 col) | 3.32 ms | 0.25 ms | **−92.5%** (13x) |
| `test_disaggregate[small]` (8 clusters x 4 col) | 2.30 ms | 0.20 ms | **−91.2%** (11x) |

The segmented path gains most because it expanded twice — segments to timesteps, then periods to the full index.

## Workflow effect

The largest in the batch, because this is the one cost that **multiplies rather than amortises**: an optimization run calls it once per solved variable.

|---|---:|---:|---:|
| `optimization_roundtrip[segments]` (50 variables) | 464.54 ms | 84.81 ms | **−81.7%** (5.5x) |
| `optimization_roundtrip[typical_periods]` (50 variables) | 189.39 ms | 39.29 ms | **−79.3%** (4.8x) |

Workflows that never call `disaggregate()` sat inside the noise floor.

<details><summary>Workflow code (<code>benchmarks/test_workflows.py</code>, added in #443)</summary>

```python
@pytest.mark.parametrize("resolution", ["typical_periods", "segments"])
def test_workflow_optimization_roundtrip(benchmark, resolution):
    """Cluster once, solve on the typical periods, expand every variable back.

    The shape of an optimization run: the solver returns one value per
    (typical period, timestep) for each of its variables, and every one of
    them has to be mapped back onto the original time index before the result
    means anything. Only the expansion repeats per variable. Segmented results
    expand through a different path than unsegmented ones, so both run here.
    """
    profiles = pd.read_csv(WIDE_CSV, index_col=0, parse_dates=True)
    profiles = pd.concat([profiles.add_suffix(f"_{i}") for i in range(5)], axis=1).iloc[
        :, :20
    ]
    segments = SegmentConfig(n_segments=12) if resolution == "segments" else None
    n_variables = 50
    benchmark.extra_info["n_variables"] = n_variables

    def run():
        result = aggregate(profiles, 36, segments=segments)
        solved_variables = result.cluster_representatives
        return [result.disaggregate(solved_variables) for _ in range(n_variables)]

```
</details>

## Measurement protocol

Arms interleaved within each repetition; **minimum** across four runs, since contention only ever adds time. Noise floor about ±5%, so nothing smaller is claimed. Workflows whose setup is `kmeans_mean` are excluded from the quoted figures — multithreaded scikit-learn makes them swing up to ~20% run to run. macOS arm64, Python 3.12, numpy 2.5.1 / pandas 3.0.5 / scikit-learn 1.8.0 / scipy 1.17.0.

## Equivalence

Full test suite green — **728 passed** at this level, including **33 new tests** (241 lines) covering the expansion directly: unsegmented and segmented, with and without extreme periods, and against externally supplied frames whose column order differs from the clustering's.

## Reviewer note

By some margin the largest change in the batch — +540/−36 across `result.py` and its tests, against +14 to +35 for the other two — and it carries the largest claim. It is also the best tested of the three. If review effort is rationed, spend it here.

### DIFF
--- a/src/tsam/result.py
+++ b/src/tsam/result.py
@@ -6,7 +6,7 @@ import math
 import warnings
 from dataclasses import dataclass, field
 from functools import cached_property
-from typing import TYPE_CHECKING, Any, cast
+from typing import TYPE_CHECKING, Any, NamedTuple, cast
 
 import numpy as np
 import pandas as pd
@@ -561,8 +561,10 @@ def _validate_disaggregate_input(
         data = data.droplevel(list(range(2, data.index.nlevels)))
 
     # Validate cluster IDs
-    data_clusters = set(data.index.get_level_values(0).unique())
-    expected_clusters = set(clustering.cluster_assignments)
+    cluster_level = data.index.get_level_values(0)
+    unique_clusters = cluster_level.unique()
+    data_clusters = set(unique_clusters)
+    expected_clusters = clustering._cluster_id_set
     if data_clusters != expected_clusters:
         missing = expected_clusters - data_clusters
         extra = data_clusters - expected_clusters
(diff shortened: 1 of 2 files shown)


FBumann:
Split into three reviewable PRs, since it bundled three independent concerns behind one 540-line diff:

- **#451** cache the cluster id set for validation — +19/−9, one file, no new tests needed
- **#452** numpy gather for period expansion — src +204/−27, tests +154
- **#453** numpy gather for segment expansion — src +76, tests +87

The dependency map made the seam obvious: the segment path shares only two guard helpers with the period path, and the validation change touches neither. Each now lands with its own attributable number instead of one combined claim — and #453 carries an explicit control showing the unsegmented benchmarks do not move.

Cumulatively identical to this PR: `test_disaggregate[wide_segmented]` 8.11 → 0.48 ms, `optimization_roundtrip[segments]` 464.5 → 84.8 ms. Verified byte-identical `src/` at the top of the split chain.

PR #454, opened 2026-07-31
feat(clustering): KMeans method config exposing restarts and a seed

FBumann:
**Draft.** Stacked on #418, which introduced the `KMedoids` method config and the `ClusterMethod` union this extends. Merge #418 first!

k-means runs `n_init` times from different random initializations and keeps the best by inertia — then draws those initializations from an **unseeded** RNG. tsam hardcodes the first at **100** (ten times scikit-learn's own default) and never sets the second. Neither is reachable from `aggregate()`.

```python
from tsam import aggregate, ClusterConfig, KMeans

aggregate(df, n_clusters=8,
          cluster=ClusterConfig(method=KMeans(n_init=10, random_state=42)))
```

The two knobs land together on purpose: `n_init` is the cost, `random_state` is what makes any of it repeatable, and **fewer restarts leave more of the outcome to the draw** — so the option that makes runs cheaper is exactly the one that most needs the option that makes them reproducible.

## Why it matters

Measured at 40 clusters over a year of 40 profiles (365 x 960 candidates):

| `n_init` | time | |
|---:|---:|---|
| **100** (tsam today) | 597 ms | |
| 10 | 59 ms | scikit-learn's default |
| 3 | 16 ms | |
| 1 | 6.3 ms | `"auto"` for k-means++ init |

Roughly linear in the restart count. The practical effect is that k-means is currently the *expensive* method: on the same energy-system workload it costs **606 ms against 37 ms for hierarchical clustering**, a 16x gap that is almost entirely restarts.

## Does setting the option change results? Yes — and here is how much

**Lowering `n_init` changes cluster assignments.** It is a different search, so it can land on a different local optimum. Anyone using the option needs to know that, so here is the measurement rather than a warning.

8760 x 48, 20 clusters, 8 runs per setting. Agreement is the adjusted Rand index between partitions — 1.0 is identical, lower is more disagreement:

| `n_init` | mean RMSE | vs 100 | agreement between two runs of *itself* | agreement with an `n_init=100` run |
|---:|---:|---:|---:|---:|
| **100** | 0.07824 | — | **0.606** | 0.632 |
| 10 | 0.07850 | +0.32% | 0.567 | 0.562 |
| 3 | 0.07910 | +1.09% | 0.518 | 0.543 |
| 1 | 0.07895 | +0.90% | 0.521 | 0.525 |

The first row is the one to read first. **At the current default, two identical calls already agree only 0.606** — because nothing is seeded, `aggregate()` is not reproducible for k-means today regardless of this PR. That is the floor everything else is measured against.

Against that floor, dropping to `n_init=10` costs **0.32% RMSE** and lands barely below the existing run-to-run spread (0.562 vs 0.606). Going to 1 costs about 1% RMSE. So the 10x speedup is close to free in fit quality, and the assignment churn it adds is smaller than the churn the library already has.

One dataset and one cluster count — enough to show the shape, not enough to prescribe a value, which is why the default is left alone and the choice handed to the caller.

## `random_state` fixes the reproducibility that was never there

The 0.606 above is not a quirk of low `n_init` — it is the status quo. Agreement across 5 runs of the same configuration:

| configuration | agreement across 5 runs |
|---|---:|
| unseeded, `n_init=100` (today's default) | 0.619 |
| unseeded, `n_init=10` | 0.592 |
| **seeded, `n_init=100`** | **1.000** |
| **seeded, `n_init=10`** | **1.000** |
| **seeded, `n_init=1`** | **1.000** |

A seed gives exact reproducibility at every restart count. Without one, `aggregate()` does not return the same clusters twice for k-means — which is worth knowing whether or not this PR lands.

`random_state=None` is the default and is precisely what scikit-learn receives today, so nothing changes for existing callers.

## With the seed fixed, what does the restart count actually buy?

Holding the seed constant makes restart count the only variable. Five seeds, mean across them:

| `n_init` | mean RMSE | vs `n_init=100` | agreement with `n_init=100` at the same seed |
|---:|---:|---:|---:|
| 1 | 0.07951 | +1.73% | 0.542 |
| 3 | 0.07905 | +1.14% | 0.558 |
| 10 | 0.07864 | +0.61% | 0.568 |
| 30 | 0.07822 | **+0.07%** | 0.589 |
| 100 | 0.07816 | — | 1.000 |

Quality improves monotonically and **flattens hard**: `n_init=30` lands within **0.07%** of `n_init=100` for a third of the work. Most of what the last 70 restarts buy is not fit.

The second column is the surprise. Even at `n_init=30`, where the fit is indistinguishable, the *partition* agrees with the 100-restart run only 0.589. Many near-equivalent optima exist, and RMSE barely separates them — so "same quality" and "same clusters" are different questions.

## The seed matters as much as the restart count

Same restart count, varying the seed:

| `n_init` | RMSE spread across 5 seeds | agreement across seeds |
|---:|---|---:|
| 1 | 0.07883 – 0.08022 | 0.488 |
| 10 | 0.07847 – 0.07876 | 0.558 |
| 30 | 0.07776 – 0.07852 | 0.585 |
| 100 | 0.07786 – 0.07845 | 0.622 |

At the current default, **switching seed moves RMSE by 0.76%** — more than going from 10 restarts to 100 (0.61%). Which seed you get is as consequential as how many restarts you buy, and today it is chosen for you at random.

That is the practical case for the pair: `random_state` decides *which* answer, `n_init` decides *how hard you look*. Both were fixed constants; neither was yours.

Caveats: one dataset, one cluster count, five seeds. Enough to show the shape and to argue the default is over-provisioned — not enough to prescribe a replacement, which is why this PR changes none.

## No change to the default

**The default is untouched, deliberately.** Fewer restarts can land on a different local optimum, so lowering it shifts cluster assignments and every golden file with them. That is a separate decision that belongs with a golden regeneration, not smuggled into an API addition. This PR only makes the knob reachable.

`n_init` defaults to `None`, meaning "whatever tsam would have used" — 100 for ordinary clustering, **30** for the duration-curve path. That sentinel is load-bearing: a plain `100` default would silently triple the restarts when `use_duration_curves=True`, which is a behaviour change disguised as a default.

## Design

Mirrors #418 exactly, so the union gains a second member rather than a second pattern:

- `ClusterMethod = ClusterMethodName | KMedoids | KMeans`
- the bare string `method="kmeans"` still works and is equivalent to `KMeans()`
- `method_name` resolves either form to `"kmeans"`; the default representation is unchanged
- `to_dict`/`from_dict` round-trip, tagged `{"type": "kmeans"}`, omitting either field when unset
- frozen and hashable, so `ClusterConfig` stays hashable

## Tests

`test/test_kmeans_config.py`, 24 tests. The ones that matter are the defaults: the bare string, an empty `KMeans()`, and the duration-curve path still reach scikit-learn with **100, 100 and 30** respectively, and an explicit value overrides each.

One test is deliberately *not* an equality assertion. tsam passes no `random_state` to scikit-learn, so k-means initialises from an unseeded RNG and two runs of either form can land on different local optima — the same nondeterminism `kmaxoids` has. The test pins that both forms take the same path and produce the same shape, and says why it cannot pin more. Worth knowing independently of this PR.

Full suite: **733 passed**. mypy back to develop's 4 pre-existing errors; ruff clean.


julian-belina:
@FBumann, this one is not marked as a draft but builts upon one that is a draft

FBumann:
Yes. It should probably also be in draft mode, but its clearly not going into main.

julian-belina (approved):

julian-belina:
Thank you very much for your contribution. I have just made a small change to the configuration objects to make them slightly more centralised.

Issue #478, opened 2026-08-21
IndexError in rescale_representatives when new_cluster empties a regular cluster


FBumann:
## Summary

`ExtremeConfig(method="new_cluster")` combined with `representation="distribution_minmax"` raises an `IndexError` inside `aggregate()`. The `new_cluster` reassignment can absorb *every* period of a regular cluster into an extreme cluster; `_count_occurrences` then omits that now-empty label, and `rescale_representatives` indexes a dense array built from the observed labels by cluster id.

This is not transfer-related — it fails in the initial aggregation, before any `apply()`. It reproduces on `develop` at 56134b0, so it predates #414.

## Repro

```python
import pandas as pd
from tsam import aggregate, ClusterConfig, ExtremeConfig

data = pd.read_csv("docs/data/testdata.csv", index_col=0, parse_dates=True)

aggregate(
    data,
    n_clusters=8,
    cluster=ClusterConfig(method="hierarchical", representation="distribution_minmax"),
    extremes=ExtremeConfig(method="new_cluster", max_value=["Load"], min_value=["T"]),
)
```

```
  File "src/tsam/pipeline/orchestrator.py", line 525, in refine_representatives
    cluster_periods_list, rescale_deviations = rescale_representatives(
  File "src/tsam/pipeline/rescale.py", line 94, in rescale_representatives
    weighting_vec[extreme_cluster_idx_arr]
IndexError: index 9 is out of bounds for axis 0 with size 9
```

## Cause

`add_extreme_periods` under `method="new_cluster"` reassigns every period that is closer to an extreme profile than to its own center (`extremes.py`). Nothing guarantees a regular cluster keeps a member, and in this config cluster 5 loses all of its periods.

`_count_occurrences` (`orchestrator.py:58`) builds its dict from `np.unique(cluster_order)`, so an emptied label is simply absent:

```python
nums, counts = np.unique(cluster_order, return_counts=True)
return {int(num): float(counts[ii]) for ii, num in enumerate(nums)}
```

`rescale_representatives` (`rescale.py:72`) then drops the labels and keeps only positions:

```python
weighting_vec = pd.Series(cluster_period_no_occur).values   # length = number of *observed* labels
...
idx_wo_peak = np.delete(np.arange(n_clusters), extreme_cluster_idx)  # indices = cluster *ids*, 0..n_clusters-1
weighting_vec[extreme_cluster_idx_arr]   # id-indexed into a position-indexed array
```

[...]

PR #479, opened 2026-08-21
fix(extremes): aggregate() raises IndexError when new_cluster empties a cluster

FBumann:
Fixes the `IndexError` in #478.

## The bug

`method="new_cluster"` reassigns every period that sits closer to an extreme profile than to its own center, and nothing stops it from taking all of them. The emptied cluster then has no entry in the occurrence counts — `_count_occurrences` builds its dict from `np.unique(cluster_order)`, so an unobserved label is simply absent.

`rescale_representatives` dropped those labels and kept only positions:

```python
weighting_vec = pd.Series(cluster_period_no_occur).values   # length = observed labels
...
weighting_vec[extreme_cluster_idx_arr]                      # indexed by cluster id
```

so `aggregate()` raised before returning:

```python
aggregate(
    data,
    n_clusters=8,
    cluster=ClusterConfig(method="hierarchical", representation="distribution_minmax"),
    extremes=ExtremeConfig(method="new_cluster", max_value=["Load"], min_value=["T"]),
)
# IndexError: index 9 is out of bounds for axis 0 with size 9
```

10 clusters, labels present `[0,1,2,3,4,6,7,8,9]`, extreme ids `8` and `9`.

## The fix

Build `weighting_vec` over `range(n_clusters)` and default an absent cluster to 0 occurrences — arithmetically what a cluster representing no periods contributes.

Nothing here is specific to `distribution_minmax`; that representation just makes the extreme profiles aggressive enough to swallow a whole cluster on `testdata.csv`. `append` never reassigns and `replace` never adds clusters, so both were always safe.

Because the highest cluster id is always `n_clusters - 1` while the array was short by the number of emptied clusters, the old code always raised rather than silently misweighting — no previously-returned result was affected by this.

## Scope

Deliberately minimal. The emptied cluster still survives as a zero-occurrence representative:

```
periods per cluster: {0: 43, 1: 58, 2: 38, 3: 63, 4: 45, 6: 31, 7: 30, 8: 15, 9: 42}
empty cluster ids: [5]
```

so `n_clusters == 10` while only 9 clusters represent anything. Whether `new_cluster` should be prevented from emptying a cluster at all — as #414's `preserve_n_clusters` path already does locally via `regular_reps` — is a design question this PR does not settle. #478 stays open for it.

## Verification

- New regression test in `test/test_extremePeriods.py`, confirmed red on `develop` (`IndexError`) and green with the fix.
- Full suite: 696 passed, 145 skipped, 5 xfailed, 1 xpassed. The xpass (`test_partial_weights_accuracy`) is pre-existing on `develop` and unrelated.
- `pandas` moved into a `TYPE_CHECKING` block since it is no longer used at runtime in this module.


FBumann:
#481 stacks on this branch and fixes the three remaining defects in #480. It also makes the line fixed here unreachable from `aggregate()` — empty clusters are dropped before rescaling runs — so this PR stays the smaller, backportable fix and #481 removes the state that produced it.

FBumann (commented):
Nice change

julian-belina:
Sorry, I accidentally introduced some failing tests. I encountered some problems with pre-commit on my machine. I have also added some more explanations and variable names to help me understand things better. I hope that doesn't interfere too much with your coding style.

Thank you for your contribution. To fully understand the issue, I examined the entire process of adding extreme periods. It would be better to be more explicit, so that we do not assume that a missing key means a zero-size cluster silently, as this can hide problems caused by upstream functions.

A more explicit way of reordering the cluster assignments, as proposed in https://github.com/FZJ-IEK3-VSA/tsam/pull/481, would be preferable. However, I think we should move the reordering function a level higher, as I found other configurations that do not include the extreme configuration and which can also result in zero-size clusters. I will discuss these separately.

Therefore, I propose changing the _count_occurrences function to create a densely populated dictionary.

julian-belina (approved):

PR #481, opened 2026-08-21
fix(extremes): transfer and cluster counts break when new_cluster empties a cluster

### LINKED ISSUE #480: Empty clusters: three further defects remain after #479
#479 fixes one of four places where an empty cluster breaks an invariant. Three remain, all reachable from the same starting point as #478: `method="new_cluster"` can absorb every member of a regular cluster, leaving a label with a representative but no periods.

> [!NOTE]
> The following content was generated by AI, while reviewing #414 against #479.

All reproductions below run on `fix/rescale-empty-cluster` (`190eec2`), i.e. **with** #479 applied.

## The state that triggers them

```python
result = aggregate(
    data, n_clusters=8, period_duration=24,
    cluster=ClusterConfig(method="hierarchical", representation="distribution_minmax"),
    extremes=ExtremeConfig(method="new_cluster", max_value=["Load"]),
)
```

```
representative ids          : [0, 1, 2, 3, 4, 5, 6, 7, 8]
labels present in assignment: [0, 1, 2, 3, 4,    6, 7, 8]
empty cluster ids           : [5]
```

Nine representatives, eight of them used. `aggregate()` now returns fine — that is #479's fix. Everything downstream of the returned object does not.

## Defect 1 — `representations()` renumbers labels densely, so a transfer loses a cluster

`use_predefined_assignments` rebuilds representatives by handing the stored `cluster_order` to `representations()`, which returns one center per *observed* label, numbered from zero. A gap in the label space is silently closed, so every stored label above the gap now points at the wrong representative — or at none.

Independent of extremes, clustering or data:

```python
import numpy as np
from tsam.algorithms.representations import representations

candidates = np.arange(6 * 4, dtype=float).reshape(6, 4)
order = np.array([0, 0, 1, 1, 3, 3])          # label 2 unused, 3 is the highest

centers, _ = representations(
    candidates, order, default="mean", representation_method="mean",
    representation_dict=None, n_timesteps_per_period=4,
)
len(centers)        # 3  -> ids 0, 1, 2
                    # stored label 3 has no representative
```

This is the root cause of defect 2 and of the `KeyError` below. Both observed effects are crashes; whether a path exists that instead reads a shifted representative without raising, I did not establish.

## Defect 2 — `rescale.py:90` repeats the pattern #479 fixed at line 94

```python
idx_wo_peak = np.delete(np.arange(n_clusters), extreme_cluster_idx)   # line 90
```

`n_clusters` here is `len(cluster_periods)`, and `extreme_cluster_idx` holds cluster *ids*. #479 fixed the same id-as-position assumption two lines down by defaulting absent clusters to zero occurrences; this line still assumes ids index into `arange(n_clusters)`.

On the fresh path it holds, because the representative list always contains every id. It breaks the moment the list is shorter than the id space — which defect 1 makes happen on transfer:

```python
result.clustering.apply(data)
# IndexError: index 8 is out of bounds for axis 0 with size 8
#   pipeline/rescale.py:90
```

Fixing defect 1 hides this, but the assumption is worth removing at the same time — it is the same latent bug, and nothing about `rescale_representatives` documents that ids must be dense.

## Defect 3 — `ClusteringResult.n_clusters` and `AggregationResult.n_clusters` disagree

Same result object, two answers:

```python
result.n_clusters              # 9  — distinct ids in cluster_representatives
result.clustering.n_clusters   # 8  — len(set(cluster_assignments))
```

`ClusteringResult.n_clusters` counts labels that were *used*; `AggregationResult.n_clusters` counts representatives that exist. They coincide only while no cluster is empty, so this is not a rounding difference — one of them is wrong for any given caller. `AggregationResult.n_clusters` at least calls its own source "authoritative" and gestures at "tsam quirks"; `ClusteringResult.n_clusters` says only "Number of clusters (typical periods)", and it is the one a user reads back after `from_json()`, with no aggregation result to compare against.

<details>
<summary>Both crash reproductions, verbatim</summary>

```
preserve_column_means=True:  IndexError: index 8 is out of bounds for axis 0 with size 8
    tsam/pipeline/rescale.py:90   idx_wo_peak = np.delete(np.arange(n_clusters), extreme_cluster_idx)

preserve_column_means=False: KeyError: '[8] not in index'
    tsam/pipeline/accuracy.py:58  reconstructed = typical_unstacked.loc[list(cluster_order)].values
```

The two differ only in whether rescaling runs first; with it off, the same missing representative surfaces one stage later in `reconstruct`.

Reproduction script:

```python
import pandas as pd
from tsam import ClusterConfig, ExtremeConfig, aggregate

raw = pd.read_csv("docs/data/testdata.csv", index_col=0, parse_dates=True)

for preserve in (True, False):
    result = aggregate(
        raw, n_clusters=8, period_duration=24, preserve_column_means=preserve,
        cluster=ClusterConfig(method="hierarchical", representation="distribution_minmax"),
        extremes=ExtremeConfig(method="new_cluster", max_value=["Load"]),
    )
    result.clustering.apply(raw)   # raises
```

`distribution_minmax` is not special — as #479 notes, it is simply the representation whose extreme profiles are aggressive enough to swallow a whole cluster on `testdata.csv`.

</details>

## Suggested order

1. Defect 1, in `representations()` or in `use_predefined_assignments` — decide whether a stored label space with gaps should be preserved (keep the empty representative) or normalised at storage time (renumber the assignment when the clustering is built). The first keeps `apply()` faithful to the run it replays; the second makes empty clusters impossible to store in the first place and would settle defect 3 too.
2. Defect 2, alongside it, so `rescale_representatives` no longer depends on dense ids.
3. Defect 3, once 1 has decided what the cluster count *means*.

#414 marks the affected combination `xfail(strict=True)` in `TestTransferExactness` with a pointer here, so the marker fails and must be removed once this is fixed.

FBumann:
Closes #480. **Stacked on #479** — based on `fix/rescale-empty-cluster`, so until that merges this PR's diff contains its commit too. Review its second commit only; the diff narrows on its own once #479 lands.

## What was wrong

All three defects in #480 come from one state: `method="new_cluster"` reassigns every period closer to an extreme profile than to its own center, and nothing stops it taking all of them. The emptied cluster kept a representative standing for no periods, leaving a hole in the label space.

Cluster ids are positions everywhere downstream — representatives live in a list indexed by id, and `representations()` returns one center per *observed* label. A hole misaligns every cluster above it. #479 stopped `aggregate()` crashing on that state; everything downstream of the returned object still broke.

I reproduced all three on `190eec2` before changing anything:

```
representative ids          : [0, 1, 2, 3, 4, 5, 6, 7, 8]
labels present in assignment: [0, 1, 2, 3, 4,    6, 7, 8]

result.n_clusters = 9   result.clustering.n_clusters = 8
apply(), preserve_column_means=True  -> IndexError: index 8 is out of bounds for axis 0 with size 8
apply(), preserve_column_means=False -> KeyError: '[8] not in index'
```

## The decision

#480 asks whether a stored label space with gaps should be **preserved** or **normalised**. This normalises, for a reason specific to what an empty cluster is: it has no members, so on transfer there is nothing to compute a representative *from*. Preserving the gap would mean faithfully replaying a representative that stands for nothing. There is no information to lose by dropping it.

So empty clusters are removed where they are created, renumbering the assignment, the extreme cluster ids and the stored center indices. Ids stay dense, and an id is always a valid position again.

The visible consequence: this configuration now returns 8 typical periods where it returned 9. That is the honest count — the ninth represented no periods and contributed nothing to the reconstruction. It also settles the question #478 left open, and makes both `n_clusters` properties agree by construction rather than by coincidence.

## Defect 2: fixed by construction, not by a guard

Dense ids make it unreachable — `rescale_representatives` can no longer be handed an extreme id without a representative. I had added a precondition there too and took it back out: it guards a state that the only remaining caller cannot produce, and #479 already made that line id-safe.

Defect 1 does keep a guard, because it stays reachable: a clustering stored *before* this change still has a hole in its label space, and `representations()` would silently renumber it — returning 3 centers for labels `[0, 1, 3]`, so cluster 3 reads cluster 2's profile. There is no representative to compute for a cluster with no members, so it now refuses the input with a sentence naming the problem instead of failing later inside numpy.

## Note on #479

After this change, the line #479 fixed is no longer reachable from `aggregate()` — empty clusters are dropped before rescaling runs. #479 is still correct and worth keeping: it is what makes `rescale_representatives` sane for a direct caller, and it is the smaller, more backportable fix of the two. But its regression test asserted that an empty cluster *survives*, which is now false, so this PR rewrites that assertion to check the cluster count instead. Worth knowing if you were planning to merge them in either order — this one depends on #479's branch but not on its fix.

## Verification

- New `test/test_empty_clusters.py`: 3 tests covering the dense label space and the two `n_clusters` agreeing, transfer under both `preserve_column_means` settings, and the `representations()` guard. All three confirmed red on `190eec2`.
- Full suite: 699 passed, 145 skipped, 5 xfailed, 1 xpassed — no regressions, no golden-data changes. The xpass (`test_partial_weights_accuracy`) is pre-existing on `develop`.
- #480 notes that #414 marks this combination `xfail(strict=True)` in `TestTransferExactness`. That marker is on #414's branch, not `develop`, so it is not removed here — it will need removing when #414 rebases.

### DIFF
--- a/src/tsam/algorithms/clustering.py
+++ b/src/tsam/algorithms/clustering.py
@@ -11,6 +11,29 @@ if TYPE_CHECKING:
     from tsam.config import ClusterMethod, Distribution, MinMaxMean
 
 
+def _densify_labels(cluster_order: np.ndarray) -> np.ndarray:
+    """Renumber cluster labels onto ``0..n-1``, closing gaps left by empty clusters.
+
+    No clustering method promises to fill every cluster it was asked for.
+    k-maxoids on data with fewer distinct period shapes than clusters is the
+    plainest case: it leaves labels nobody uses, so the assignment comes back as
+    e.g. ``[0, 1, 3]`` for ``n_clusters=8``.
+
+    Cluster ids are positions everywhere downstream — representatives live in a
+    list indexed by id, and
+    :func:`~tsam.algorithms.representations.representations` returns one center
+    per *observed* label — so an unused label shifts every cluster above it and
+    leaves the highest one without a representative. Nothing is lost by dropping
+    it: a cluster with no members has no periods to compute a representative
+    from, and contributes nothing to the reconstruction.
+
+    Renumbering preserves order, so a label space that is already dense — every
+    method's normal outcome — comes back with its labels unchanged.
+    """
+    labels = np.unique(cluster_order)
+    return np.asarray(np.searchsorted(labels, cluster_order), dtype=int)
+
+
 def assign_clusters(
     candidates: np.ndarray,
     n_clusters: int,
@@ -27,7 +50,28 @@ def assign_clusters(
 
     Valid ``cluster_method`` values: 'averaging', 'kmeans', 'kmedoids',
     'kmaxoids', 'hierarchical', 'contiguous'.
+
+    Returns:
+        The cluster id of each period, as a dense label space: the ids run
+        ``0..n-1`` with no gaps, where ``n`` is the number of clusters that
+        actually received periods. That can be fewer than *n_clusters* — see
+        :func:`_densify_labels` for why the gaps are closed rather than kept.
     """
+    return _densify_labels(
+        _raw_assignment(
+            candidates, n_clusters, cluster_method, n_iter=n_iter, solver=solver
+        )
+    )
+
+
+def _raw_assignment(
+    candidates: np.ndarray,
+    n_clusters: int,
+    cluster_method: str,
+    n_iter: int = 100,
+    solver: str = "highs",
+) -> np.ndarray:
+    """Dispatch to the clustering method, returning its labels as it produced them."""
     if cluster_method == "averaging":
         n_sets = len(candidates)
         cluster_size = n_sets // n_clusters
--- a/src/tsam/algorithms/clustering.py
+++ b/src/tsam/algorithms/clustering.py
@@ -61,11 +61,14 @@ def assign_clusters(
         actually received periods. That can be fewer than *n_clusters* — see
         :func:`densify_labels` for why the gaps are closed rather than kept.
     """
-    return densify_labels(
-        _raw_assignment(
+    raw_cluster_order=_raw_assignment(
             candidates, n_clusters, cluster_method, n_iter=n_iter, solver=solver
         )
-    )
+    # Clustering configuration might yield non-dense cluster indices (e.g., when
+    # fewer clusters are needed due to duplicated periods). We reassign indices
+    # to be contiguous so downstream functions have a clear, well-defined data structure.
+    cluster_order= densify_labels(cluster_order=raw_cluster_order)
+    return cluster_order
 
 
 def _raw_assignment(
--- a/src/tsam/algorithms/clustering.py
+++ b/src/tsam/algorithms/clustering.py
@@ -20,12 +20,13 @@ def densify_labels(cluster_order: np.ndarray) -> np.ndarray:
     e.g. ``[0, 1, 3]`` for ``n_clusters=8``.
 
     Cluster ids are positions everywhere downstream — representatives live in a
-    list indexed by id, and
-    :func:`~tsam.algorithms.representations.representations` returns one center
-    per *observed* label — so an unused label shifts every cluster above it and
-    leaves the highest one without a representative. Nothing is lost by dropping
-    it: a cluster with no members has no periods to compute a representative
-    from, and contributes nothing to the reconstruction.
+    list indexed by id — so an unused label puts every cluster above it out of
+    step with its representative.
+    :func:`~tsam.algorithms.representations.representations` refuses such a
+    label space outright, which is why densifying happens here, before it is
+    called. Nothing is lost by dropping the label: a cluster with no members has
+    no periods to compute a representative from, and contributes nothing to the
+    reconstruction.
 
     Renumbering preserves order, so a label space that is already dense — every
     method's normal outcome — comes back with its labels unchanged.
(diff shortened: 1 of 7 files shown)


julian-belina:
This change is generally good. However, during my previous review, I noticed that it does not go far enough. There are more cases that can result in sparsely populated cluster_order matrices. It should always run, not only when extreme cases are configured.

Here is one example:
```
"""Empty clusters without any ExtremeConfig — reachable from clustering alone.

Twenty days, only three distinct day shapes. k-maxoids cannot fill eight
clusters, so it leaves holes in the label space. No extremes are configured
anywhere in this script.
"""

import numpy as np
import pandas as pd

import tsam.pipeline.orchestrator as orch
import tsam.pipeline.rescale as rescale_mod
from tsam import ClusterConfig, aggregate

# --- spy on what rescale_representatives is handed -------------------------
_real_rescale = rescale_mod.rescale_representatives


def spy(**kw):
    centers = kw["cluster_periods"]
    counts = kw["cluster_period_no_occur"]
    n = len(centers)
    print(f"    len(cluster_periods)   = {n}   <- one center per OBSERVED label")
    print(f"    cluster_counts keys    = {sorted(counts)}   <- the sparse labels")
    print(f"    weighting_vec (>=#479) = {[float(counts.get(i, 0.0)) for i in range(n)]}")
    print(f"    weighting_vec (pre-479)= {[float(v) for v in pd.Series(counts).values]}")
    return _real_rescale(**kw)


orch.rescale_representatives = spy

# --- data: 20 days, 3 distinct shapes --------------------------------------
index = pd.date_range("2020-01-01", periods=24 * 20, freq="h")
values = np.concatenate([[1.0, 5.0, 9.0][i % 3] * np.ones(24) for i in range(20)])
data = pd.DataFrame({"x": values, "y": values * 2}, index=index)

np.random.seed(0)  # k-maxoids uses numpy.random; pin it so this reproduces

print("aggregate(n_clusters=8, kmaxoids, representation='mean'), no extremes:")
try:
    result = aggregate(
        data,
        n_clusters=8,
        cluster=ClusterConfig(method="kmaxoids", representation="mean"),
    )
    print(f"  -> completed, n_clusters={result.n_clusters}")
except Exception as exc:  # noqa: BLE001
    print(f"  -> {type(exc).__name__}: {exc}")

```



```
aggregate(n_clusters=8, kmaxoids, representation='mean'), no extremes:
    len(cluster_periods)   = 3   <- one center per OBSERVED label
    cluster_counts keys    = [0, 1, 3]   <- the sparse labels
    weighting_vec (>=#479) = [7.0, 7.0, 0.0]
    weighting_vec (pre-479)= [7.0, 7.0, 6.0]
  -> KeyError: '[3] not in index'

```

julian-belina (approved):

julian-belina:
I think this should be OK. In my opinion, keeping dense indices makes it easier to detect bugs further down the line. However, if you can think of an easier way, I'm open to it, @FBumann.

